PR #169, opened 2015-08-17
Query a case-insensitive match for username

safwanrahman:
Some persons username maybe in lower or upper case. This would fix the issue.
Like, there is user named "topal" But, trying to get profile "Topal" return user not found. But this can be avoid by a case-insensitive DB query.

r?

### DIFF
--- a/standup/apps/status/views.py
+++ b/standup/apps/status/views.py
@@ -94,7 +94,7 @@ def user(slug):
     """The user page. Shows a user's statuses."""
     db = get_session(current_app)
 
-    user = db.query(User).filter_by(slug=slug).first()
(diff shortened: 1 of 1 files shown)


safwanrahman:
@rehandalal r?

willkg:
In order to do this, names should be unique case-insentiviely in the db otherwise it's possible to have Topal and topal users, but then the above changes means you can only ever access one of them.

This should have some tests, too.

Travis is failing the test suite, but I think it's a problem with dependencies or something. I don't think it's related to this change.

willkg:
I see the use case here, but this implementation doesn't work right. I think I'm going to close this out.

If you're still interested in this, it's worth writing up an issue for it.

PR #192, opened 2016-08-27
Api bits

willkg:
Some mild scaffolding for api stuff.

### DIFF
--- a/standup/api/tests/test_views.py
+++ b/standup/api/tests/test_views.py
@@ -1,0 +1,235 @@
+import json
+
+from django.core.urlresolvers import reverse
+from django.utils.encoding import force_bytes
+from django.test import Client, TestCase
+
+from standup.api.tests.factories import SystemTokenFactory
+from standup.status.models import Project, Status
+from standup.status.tests.factories import ProjectFactory, StatusFactory
+from standup.user.tests.factories import StandupUserFactory
+
+
+def _content_generator():
+    """Generator for Jonathan Coulton songs"""
+    content = [
+        'Baby Got Back',
+        'Chiron Beta Prime',
+        'Christmas with You Is the Best',
+        'Code Monkey',
+        'I Crush Everything',
+        'I Say Arrr, You Say Woo...',
+        'Re: Your Brains',
+        'Redshirt',
+        'Shop Vac',
+        'Skullcrusher Mountain',
+        'Stroller Town',
+        'The Presidents',
+        'Todd the T1000',
+        'When I\'m 25 or 64',
+    ]
+    while True:
+        for song_title in content:
+            yield song_title
+
+
+content_generator = _content_generator()
+
+
+class JSONClient(Client):
+    """Django client that provides a .json_post() method
+
+    The ``.json_post()`` method makes it easier to test API endpoints because it deals with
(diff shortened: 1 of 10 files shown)


willkg:
@pmac What do you think of this? It's just an example and it's missing some bits, but the rough sketch is there.

There are two nice things about this:

1. the view code is almost identical to what's in the current Standup--the only changes are converting SQLAlchemy things to Django ORM things; that'll reduce the work in rewriting things
2. it's pretty much like how Django View classes work, so it's not something unicorn-y

I might create a JSONHttpResponse class just to make that feel a bit better.

pmac:
Like [this](https://github.com/mozilla/bedrock/blob/8fccef92d608b24aa4aa1c7f40074a614a5707f3/bedrock/mozorg/util.py#L27-L27)?

standup/api/views.py
```diff
@@ -0,0 +1,152 @@
...
+    """Authentication exception"""
+    pass
+
+
+class APIView(View):
+    """API view that looks at the world in JSON"""
+    @classmethod
+    def as_view(cls, *args, **kwargs):
+        # API calls use token authentication and don't need csrf bits
+        return csrf_exempt(super(APIView, cls).as_view(*args, **kwargs))
```
pmac: If you'd rather I think you can just decorate `dispatch` with `csrf_exempt` and not need to override `as_view` at all, since you're overriding `dispatch` below anyway.

pmac:
I made one more comment, but I like this code a lot. Let's get it in and keep iterating. r+

willkg:
Bunch of fixes to the APIView. Amongst other things, this one passes some tests!

Still needs some tests regarding missing data and some other things, but it's pretty close.

willkg:
^^^ That's got enough API infrastructure to write an API endpoint that should work with standup-irc.

@pmac r? when you gots time?

standup/api/tests/test_views.py
```diff
@@ -0,0 +1,235 @@
...
+from django.test import Client, TestCase
+
+from standup.api.tests.factories import SystemTokenFactory
+from standup.status.models import Project, Status
+from standup.status.tests.factories import ProjectFactory, StatusFactory
+from standup.user.tests.factories import StandupUserFactory
+
+
+def _content_generator():
+    """Generator for Jonathan Coulton songs"""
```
pmac: Great. Now I have to listen to all of these right now. HAVE TO!! 🎧

standup/api/tests/test_views.py
```diff
@@ -0,0 +1,235 @@
...
+    content = [
+        'Baby Got Back',
+        'Chiron Beta Prime',
+        'Christmas with You Is the Best',
+        'Code Monkey',
+        'I Crush Everything',
+        'I Say Arrr, You Say Woo...',
+        'Re: Your Brains',
```
pmac: I mean, no one's gonna eat your eyes.

pmac:
So full of win I can't stand it. r+ :shipit:

standup/api/views.py
willkg: I like that idea. I'll test it out. Plus I should write a test to make sure it's working csrf-wise.

PR #195, opened 2016-08-29
Improve docker env. Fix staticfiles.

pmac:
Move to new docker process model. This is a large change, but I think it's good. Allows us some flexibility and a decent local dev docker environment as well as prod docker build potential via the new Heroku stuff. Let me know what you think.


docker-compose.yml
```diff
@@ -1,25 +1,44 @@
...
+      - "8000:8000"
+    volumes:
+      - .:/app
+    links:
+      - db
+    environment:
+      - DATABASE_URL=postgres://postgres@db/postgres
+      - DEBUG=False
+      - ALLOWED_HOSTS=localhost,127.0.0.1,
+      - SECRET_KEY=59114b6a-2858-4caf-8878-482a24ee9542
```
willkg: Pretend I don't understand this part. Seems like we're putting a secret key into git for a prod environment. How do we expect to run prod?

willkg:
I'm so glad you're working on this.

My only curiosity is whether the changes in this affect what we have in the README. Do we need to update anything there?

docker-compose.yml
pmac: Not planning on running this as the prod environment. In fact docker-compose won't be used in prod. This is meant as a way to try running the site locally but in a prod-like environment. It should be documented as such though. And it might be better to use a non-official-looking key here: something like `ITSASEKRIT` that is obviously a placeholder?

pmac:
Good catch on the README. I'll update.

pmac:
Hmmm... Actually I think the README is okay still. The Heroku bit will change, but hasn't yet. I've changed how a bunch of that works behind the scenes, but the developer interface with `make` remains unchanged (which I like a lot btw).

willkg:
It might be "better" to have fake-looking `SECRET_KEY` values, but it's a feel-good thing and not something that affects anything as far as I can imagine.

Awesome! Thank you sooooo much for doing this.

PR #196, opened 2016-08-29
Fix local testing via docker-compose

pmac:
Oops. A couple more things that I didn't quite get in before the last PR merged. Thanks for looking BTW.

### DIFF
--- a/bin/run-common.sh
+++ b/bin/run-common.sh
@@ -2,4 +2,4 @@
 
 # Migrate the db
 python3 manage.py migrate --noinput
-python3 manage.py collectstatic --noinput -c
+python3 manage.py collectstatic --noinput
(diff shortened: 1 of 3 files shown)


pmac:
@willkg for future reference: this is the kind of minor change I'd be comfortable just pushing to master w/o r? if you are. If you'd rather we take a look at every change I'm also fine with that. In this pre-release phase though I am usually a little more lax. Opinions?

willkg:
Oops--sorry about that!

Standup doesn't work right now, so landing minor stuff directly is fine.  Docs fixes and typos and other minor things seems fair game for landing directly. We should throw a PR together for major changes that we want a second set of eyes on. I think at this stage PR is less for approval to land and more a request for a second set of eyes or a way to communicate things we're wondering about.

Does that sound ok to you?

pmac:
That is my feeling exactly. Thanks! 👍

pmac:
I'll just push this commit to master then.

Issue #180, opened 2016-08-26
figure out how to migrate users from Persona to GitHub auth


willkg:
After we implement Github auth and possibly as part of the migration script, we're going to need to figure out how to migrate users from Persona to Github auth.

This issue involves:

1. getting a backup of the db
2. analyzing the db to figure out what the situation is
3. figuring out a plan

As a side note, we should only be migrating users who have used Standup in the last x months.

pmac:
We could potentially leave this to the users. If we started with support for both Persona and Github we could prompt those signed in but who don't have a github account associated to add one. This should catch most of the active users I think. After we remove Persona when it is shut down we could fall back to email matching or prompting users to file issues if they've lost access to their account. It's possible we could also match with the existing "github_handle" field, but that is in no way verified AFAIK, and thus could be a security issue.

pmac:
So are we kinda done with this one? We've got that new login page, and if a user logs in using Persona they'll be annoyed with a banner at the top of every page asking them to add their Github account. I'm open to other suggestions, but I think we might be "good enough".

willkg:
I think that covers everything. Yay team!

PR #197, opened 2016-08-30
Add db migrations for legacy database

### LINKED ISSUE #178: write db migration script
When we go to production with the new Standup, we're going to need to migrate the data from the old system to the new system.

I'm pretty sure the two systems don't share any table names. Maybe we can run `./manage.py migrate` which sets up all the Django tables and then run a script that copies all the data from the old tables to the new tables in the same database. Then later wipe out the old tables.

This issue covers figuring out how we want to do that, writing the relevant script, pulling a backup of the db locally and testing it all out.

pmac:
* Improve Django admin
* Enable security settings
* Changes to docker required for Heroku

Fix #178

### DIFF
--- a/docker/standup_base
+++ b/docker/standup_base
@@ -27,5 +27,5 @@ RUN pip3 install --require-hashes --no-cache-dir -r requirements.txt
 
 COPY . /app
 
-RUN adduser --uid 1000 --disabled-password --gecos '' --no-create-home webdev
-RUN chown webdev.webdev -R .
+# process static files
+RUN bin/phb-manage.sh collectstatic --noinput
--- a/standup/api/admin.py
+++ b/standup/api/admin.py
@@ -5,4 +5,8 @@ from .models import SystemToken
 
 @admin.register(SystemToken)
 class SystemTokenAdmin(admin.ModelAdmin):
-    pass
+    list_display = ('token', 'summary', 'enabled', 'contact', 'created')
+    list_editable = ('enabled',)
+    list_filter = ('enabled',)
+    readonly_fields = ('token',)
+    ordering = ('created',)
--- a/standup/status/models.py
+++ b/standup/status/models.py
@@ -12,7 +12,13 @@ class Project(models.Model):
     name = models.CharField(max_length=100, unique=True)
     slug = models.CharField(max_length=100, unique=True)
     color = models.CharField(max_length=6)
-    repo_url = models.URLField()
+    repo_url = models.URLField(max_length=100)
+
+    class Meta:
+        db_table = 'project'
+
+    def __str__(self):
+        return self.name
 
     def __repr__(self):
         return '<Project: [%s] %s>' % (self.slug, self.name)
@@ -38,6 +44,12 @@ class Status(models.Model):
     reply_to = models.ForeignKey(
         'self', blank=True, null=True, default=None,
         on_delete=models.SET_DEFAULT)
+
+    class Meta:
+        db_table = 'status'
+
+    def __str__(self):
+        return 'Status from %s' % self.user.username
 
     def __repr__(self):
         return '<Status: %s: %s>' % (self.user.username, self.content)
--- a/standup/user/migrations/0003_legacy_db_move.py
+++ b/standup/user/migrations/0003_legacy_db_move.py
@@ -1,0 +1,41 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+def migrate_legacy_users(apps, schema_editor):
+    LegacyUser = apps.get_model('user', 'LegacyUser')
+    StandupUser = apps.get_model('user', 'StandupUser')
+    User = apps.get_model('auth', 'User')
+    for old_user in LegacyUser.objects.all():
+        names = old_user.name.split(maxsplit=1)
+        new_user = User.objects.create(
+            username=old_user.username,
+            first_name=names[0],
+            last_name=names[1] if len(names) > 1 else '',
+            email=old_user.email or '',
+            is_staff=old_user.is_admin,
+            is_superuser=old_user.is_admin,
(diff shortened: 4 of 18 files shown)


pmac:
To try this locally you'll need to load a dump of the prod db into your local. I can get a procedure together for this if you want to do it, or we can just see it working in the staging env I'll have up shortly after this merges.

docker/standup_base
```diff
@@ -26,6 +26,4 @@ COPY requirements.txt /app/requirements.txt
 RUN pip3 install --require-hashes --no-cache-dir -r requirements.txt
 
 COPY . /app
-
-RUN adduser --uid 1000 --disabled-password --gecos '' --no-create-home webdev
-RUN chown webdev.webdev -R .
+RUN DEBUG=False SECRET_KEY=foo DATABASE_URL=sqlite:// ./manage.py collectstatic --noinput
```
willkg: I wonder if there's a better way to run collectstatic other than setting these things in the environment. Are there other management commands we'd want to run ignoring these variables, too? Should we make a bash wrapper for it? Something like `phb_manage.sh`:

```
#!/bin/bash

# phb_manage.sh runs manage.py with oblivious hand-wavey default settings
# suitable for management commands that will work despite the wildly ignorant
# nature of hand-wavey default settings.
#
# Usage: phb_manage.sh [COMMAND] [ARGS]
#
# Example: phb_manage.sh collectstatic --noinput

DEBUG=False
SECRET_KEY=foo
DATABASE_URL=sqlite://

./manage.py $@
```

Then we can use it elsewhere and it's probably easier to remember to update maybe? I'm just thinking out loud.

standup/api/admin.py
```diff
@@ -5,4 +5,8 @@
 
 @admin.register(SystemToken)
 class SystemTokenAdmin(admin.ModelAdmin):
-    pass
+    list_display = ('token', 'summary', 'enabled', 'contact', 'created')
+    list_editable = ('enabled',)
+    list_filter = ('enabled',)
+    readonly_fields = ('token',)
+    ordering = ('created',)
```
willkg: Thank you!

For the record, I haven't actually tested any of the admin things I've thrown together figuring we'd fix it when someone looked at it with some knowledge of what they wanted to do there.

standup/status/models.py
```diff
@@ -12,7 +12,13 @@ class Project(models.Model):
     name = models.CharField(max_length=100, unique=True)
     slug = models.CharField(max_length=100, unique=True)
     color = models.CharField(max_length=6)
-    repo_url = models.URLField()
+    repo_url = models.URLField(max_length=100)
+
+    class Meta:
+        db_table = 'project'
```
willkg: Mmm... Part of me likes this migration idea. But the other part of me is thinking about maintenance in the long term and we're going to end up with db tables that have names that have historical baggage.

What do you think? Is it "easy" to rename these tables later on? Am I overthinking this?

standup/status/models.py
```diff
@@ -12,7 +12,13 @@ class Project(models.Model):
...
+
+    def __str__(self):
+        return self.name
```
willkg: Feel free to fix all my `__str__` things. I can't remember what Django does with them, so they're all probably wrong.

willkg: Wrong or missing apparently. Ditto for `__repr__`.

standup/user/migrations/0003_legacy_db_move.py
```diff
@@ -0,0 +1,41 @@
...
+        ghh = old_user.github_handle
+        if ghh.startswith('https://github.com/'):
+            ghh = ghh[19:]
+
+        StandupUser.objects.create(
+            id=old_user.id,
+            user=new_user,
+            slug=old_user.slug,
+            github_handle=ghh,
+        )
```
willkg: I wonder what else `django_allauth` with the github auth set up will require that we'll also need to migrate. When we land this, we should toss a comment into issue #180 to tweak this code if need be.

willkg:
I had some thoughts. I don't think any of them matter a whole lot. I didn't test the migration against the current prod db.

I think this is good enough to land now and we can hone things later when we're doing more methodical testing of everything including switching from the old system to the new one.

I'll let you make that call and merge it accordingly as you so desire given that I'm on PTO and not really here. :p

standup/api/admin.py
pmac: Yeah. I just did a quick initial fleshing-out pass. I'm sure we'll want more tweaks, but I also wanted to see something nicer than default in my local testing :)

docker/standup_base
pmac: I like that. I don't like the ugly command in the docker file, but at build time there are no env vars, and doing the collectstatic step at run time in `run-prod.sh` isn't optimal since that delays the starting of the server (which will likely get worse if we have a sass processing or minification step). I can't immediately think of another management command we'd need to run, but it won't hurt to have the option. I'll add this. Thanks 🍰

standup/status/models.py
pmac: The current names are pretty reasonable and should not conflict with future names that Django will use. I'm thinking we just leave it, but I could be convinced otherwise. I'm pretty sure I could write a migration that would pretty easily just rename the tables; I'm just not sure there's much point.

standup/status/models.py
pmac: I think `__str__` is for display in the admin, and `__repr__` in the REPL. I mostly added them when I first looked at the admin with the ported prod data and wanted to see more interesting names than "Project instance" or whatever.

standup/user/migrations/0003_legacy_db_move.py
pmac: We should. I'm unsure what if any difference will be between `username` and `slug`. Perhaps the former will end up being some github auth ID? Dunno. We'll probably have to rework some of this when we do that, I agree.

PR #201, opened 2016-08-30
Fix #173: Use default staticfiles storage for tests

### LINKED ISSUE #173: "make test" needs some tlc
Right now, running `make test` runs collectstatic and then runs the tests. collectstatic is required because it puts all the static files in the place where whitenoise can find them. This is needed because they're referred to in the templates.

Is there a way to run the tests without running collectstatic every time? Should we do some make-fu build dependency stuff to make that happen? Can we accurately figure out when we need to re-run collectstatic?

Further, if you want to run a single test, you have to emulate what `make test` does. I think this probably means we want to move the test execution into a bin/test.sh file which can pass along arguments and then have `make test` run that. I think that's what Erik suggested when I asked him about similar things with DXR.

Anyhow, this issue covers making `make test` experience better and more helpful for real-world development.

pmac:
It's actually the Gzip Manifest staticfiles storage that barfs if the files are missing. I think this is a fine fix for this (we also do this in bedrock) since I don't think static files should be required for python unittests, but I'm happy to hear a differing opinion.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,27 @@ env:
     - HEROKU_PROD_APP=standups
     - HEROKU_STAGE_APP=standupstage
     - HEROKU_PROC_TYPE=web
+    - DOCKER_CACHE_FILE=/home/travis/docker/cache.tar.gz
+cache:
+  directories:
+    - /home/travis/docker/
 before_install:
   - docker -v
   - docker-compose -v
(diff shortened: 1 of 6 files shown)


pmac:
I'll go ahead and merge this, but reopen the issue for further discussion if we'd like to fix it a different way.

willkg:
This looks fine to me. r+

PR #205, opened 2016-09-06
Fix #179: Add Github Auth

### LINKED ISSUE #179: implement github auth
@pmac and I looked through the relevant standup issues and talked to some people and then mulled over our options for auth now that Persona is going away. We settled on this set:

1. nopassword: the entire thing is email-driven
2. google auth
3. github auth

Seems non-trivial to get email set up for a variety of reasons. Further, with SUMO we constantly had problems with email not getting to the recipient and not having good ways to diagnose what happened. That was irritating as all hell. While I like the nopassword method in theory, I'm inclined to not do this because of the implementation and maintenance effort I think will be required.

I find google auth irritating. But it's an option.

Github auth seems the easiest of the three to implement and maintain, so I think we're going to go with this one for now. We're going to use django-allauth which MDN is using so we can expand into other social providers later if need be.

This issue covers implementing Github auth for the MVP for now.

pmac:
Currently will lookup user based on email address. So if the email is already in the system they'll be associated with the Github account. This is only okay since we've got verified emails from Persona currently in the db, and we assume the email has been verified by Github.

TODO verify the assumption that Github is returning only verified email addresses.


pmac:
@willkg this works for me! Ping me for the keys required for the settings to make this work for you locally.

pmac:
@willkg would you rather merge this one as is, or for me to go ahead and add the login page we discussed first?

willkg:
Mmm... I vote land this now and then do the login view in a subsequent PR.

I skimmed the code and my head is mushy, so about all I got out of skimming it is the following series of thoughts:

1. Oh. BrowserID. How I miss you.
2. I see Python.
3. Look--pretty sprites!
4. I never really though about how much I like this shade of mint green in the PR diffs on GitHub. It's nice.

Given that, I vote land it and we can figure out issues later. I'll write up an issue for testing out the new standup and we can use that as a test plan and dump anything else we know needs some quality testing there.

pmac:
I'm with you. I know this is a big one, and my brain is mushy thinking about it all again now as well :(

Imma give it a quick once over again and merge. Thanks!

PR #207, opened 2016-09-07
More api

willkg:
This implements the remaining required API endpoints that we need for the standup-irc bot.

Two caveats:

1. I didn't implement the ability for admin to change user settings via the api--they should do that in the webapp.
2. I didn't implement any of the team management stuff which is /api/v2 endpoints that are all pretty different. I wrote some notes on that, but I kind of think we should do that in the webapp, too. Maybe later we can implement a new api for it or something.

Thoughts? r?

### DIFF
--- a/standup/api/views.py
+++ b/standup/api/views.py
@@ -140,7 +140,7 @@ class AuthenticatedAPIView(APIView):
         request.auth_token = token
 
 
-class StatusPost(AuthenticatedAPIView):
+class StatusCreate(AuthenticatedAPIView):
     def post(self, request):
         token = request.auth_token
 
--- a/standup/api/views.py
+++ b/standup/api/views.py
@@ -155,7 +155,7 @@ class StatusPost(AuthenticatedAPIView):
 
         # Validate we have the required fields.
         if not (username and content):
-            return HttpResponseBadRequest('Missing required fields')
+            return HttpResponseBadRequest('Missing required fields.')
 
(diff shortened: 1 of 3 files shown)


standup/api/views.py
```diff
@@ -190,3 +190,50 @@ def post(self, request):
...
+        return HttpResponseJSON({'id': status_id})
+
+
+class UpdateUser(AuthenticatedAPIView):
+    def post(self, request, username):
+        token = request.auth_token
+
+        # FIXME: Authorize this operation.
+
+        user = StandupUser.objects.filter(user__username=username).first()
```
pmac: I think this should search based on `slug`. Currently the Profile page uses the `slug` field for "IRC Handle".

pmac:
This LGTM 👍 Just that one comment about `slug` vs. `username`. Once you're satisfied I say merge.

standup/api/views.py
willkg: Mmm... So, the old code for the API is here:

https://github.com/mozilla/standup/blob/36120b10d6906d174f414e6766272ba25c3386bd/standup/apps/api/views.py#L218

I think I want to leave this as is because that's how it "works" now, but we should fix it when we work on that "wtf irc nick, slug, username, ..." bug.

willkg:
I tossed a note in the issue about the username=ircnick thing:

https://github.com/mozilla/standup/issues/212#issuecomment-245627718

I'm going to merge the rest of this now.

PR #214, opened 2016-09-07
Remove the help page and add setting for help URL

pmac:
If the setting isn't set there just isn't a "Help & FAQ" link.


pmac:
@willkg r?

willkg:
Looks good! r+

PR #215, opened 2016-09-08
Prep for SSL

pmac:
We've filed a but to get an SSL cert. This gets us ready for turning that on.

### DIFF
--- a/standup/settings.py
+++ b/standup/settings.py
@@ -144,11 +144,15 @@ USE_TZ = True
 SECURE_BROWSER_XSS_FILTER = config('SECURE_BROWSER_XSS_FILTER', default='true', parser=bool)
 # this should be 31536000 in prod (1 year)
 SECURE_HSTS_SECONDS = config('SECURE_HSTS_SECONDS', default='0', parser=int)
+SECURE_HSTS_INCLUDE_SUBDOMAINS = config('SECURE_HSTS_INCLUDE_SUBDOMAINS',
+                                        default='false', parser=bool)
 SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
 SECURE_SSL_REDIRECT = config('SECURE_SSL_REDIRECT', default='false', parser=bool)
+SESSION_COOKIE_SECURE = CSRF_COOKIE_SECURE = SECURE_SSL_REDIRECT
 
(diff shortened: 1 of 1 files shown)


pmac:
@willkg I'm not sure if linking directly to the default gravatar on github is the right call, but I also don't think linking directly to standu.ps is right either, and it has to be a fully qualified public url. I'm open to suggestions, or even just going with [the default "mm"](https://github.com/twaddington/django-gravatar#configuring).

standup/settings.py
```diff
@@ -144,11 +144,17 @@ def path(*paths):
...
+                                        default='false', parser=bool)
 SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
 SECURE_SSL_REDIRECT = config('SECURE_SSL_REDIRECT', default='false', parser=bool)
+SESSION_COOKIE_SECURE = CSRF_COOKIE_SECURE = SECURE_SSL_REDIRECT
 
-GRAVATAR_DEFAULT_IMAGE = 'http://www.standu.ps/static/img/default-avatar.png'
-GRAVATAR_DEFAULT_SECURE = False
+# must be a full publically accessible URL, or a name of a gravatar default theme
+GRAVATAR_DEFAULT_IMAGE = 'https://raw.githubusercontent.com/mozilla/standup/master/standup/' \
+                         'status/static/img/default-avatar.png'
```
willkg: So... this is a python file and if we know by this point whether the site is https and hsts, then we can know which default url to use.

Oh, wait... Do we not know the protocol and host for this instance, yet? If not, we could add that as a setting and then key off of that. Seems like it'd help to know it possibly in other places, too.

pmac: The problem is less that we don't know the protocol, and more that we need a public url here. If we don't have a public url and use the current host this would never work locally e.g.. The way it works is that the default URL is appended to the gravatar URL as a query param, which I guess the gravatar service loads and serves via their URL. So the protocol doesn't affect the security of the site, but it would break if we introduce an SSL redirect for prod if we leave the old default.

pmac: Perhaps the best thing really is just to use "mm" or one of the gravatar defaults and configure the URL for each instance.

pmac:
Updated.

willkg:
Looks good to me! r+

Issue #224, opened 2016-09-20
[meta] Get higher score on Observatory


freddyb:
https://observatory.mozilla.org/analyze.html?host=standu.ps

- [ ] CSP #220
- [ ] HSTS #221 
- [ ] HTTP to HTTPS redirect #222
- [ ] Security headers (XFO et al.)  #223

freddyb:
To be fair, this isn't about the Observatory score, but rather about best practices in web security :)

pmac:
Most of this is predicated on getting the cert, which according to [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1300770#c1) is in progress.

pmac:
Which we're tracking in #168.

freddyb:
Yes. To be clear, this isn't me complaining. I'm happy to help do those things via pull requests. :)

pmac:
We're up to an `A`! https://observatory.mozilla.org/analyze.html?host=www.standu.ps

pmac:
I'll be bumping the HSTS header value to 1 year later this week. Just want to give it some time before locking us in to discover any issues.

PR #225, opened 2016-09-20
Add XFrameOptionsMiddleware (see #224)

freddyb:
This sets the X-Frame-Options header to SAMEORIGIN for every outgoing HttpResponse according to the [django docs](https://docs.djangoproject.com/en/1.9/ref/clickjacking/#how-to-use-it)

### DIFF
--- a/standup/settings.py
+++ b/standup/settings.py
@@ -60,6 +60,7 @@ MIDDLEWARE_CLASSES = (
     'whitenoise.middleware.WhiteNoiseMiddleware',
     'standup.status.middleware.NewUserProfileMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
+    'django.middleware.clickjacking.XFrameOptionsMiddleware',    
     'django.middleware.common.CommonMiddleware',
     'django.middleware.csrf.CsrfViewMiddleware',
     'django.contrib.auth.middleware.AuthenticationMiddleware',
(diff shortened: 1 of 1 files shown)


willkg:
This should get a test so that we know it's working properly and also get a heads-up if we break it accidentally at some later point. Pretty sure I did this with fjord and/or kitsune, so we could grab it from there. I'll look for the relevant code later.

freddyb:
Agreed. Mostly for regression testing, I suppose.
I don't expect a lot of good reasons for putting standu.ps in a frame.

pmac:
Unless someone has added it to a big-screen at an office via [Corsica](https://github.com/mozilla/moz-corsica/). I don't think that's happened, and I guess we'll hear about it if it has :)

mythmon:
Mozilla's Corsica config is [here](https://github.com/mozilla/moz-corsica/blob/master/state.json). I don't see any Standups urls there.

pmac:
There you have it. Thanks @mythmon.

willkg:
@pmac Looks like @freddyb is doing the patches by editing in github and doesn't have a dev environment working, so if this looks good to you, land it and I'll throw a test together later.

pmac:
This is live in prod, but we can leave this open to remind us to write the test.

willkg:
PR #253 adds the above test.

Issue #220, opened 2016-09-20
Implement Content-Security-Policy


freddyb:
Standup should implement Content-Security-Policy (CSP), as a safeguard against XSS.

PR #226, opened 2016-09-20
This should add the X-Content-Type: nosniff header

freddyb:
On a side-note, all configs default to either 'true' or 'false', which I found _slightly_ confusing.

I suppose 'True' and 'False' (still quoted, by uppercase) would be a bit more pythonic. 

I'll file another PR should we care enough to change this (cc @pmac to decide this :))

### DIFF
--- a/standup/settings.py
+++ b/standup/settings.py
@@ -144,7 +144,7 @@ USE_L10N = True
 USE_TZ = True
 
 # security
-
+SECURE_CONTENT_TYPE_NOSNIFF = config('SECURE_CONTENT_TYPE_NOSNIFF', default='true', parser=bool)
 SECURE_BROWSER_XSS_FILTER = config('SECURE_BROWSER_XSS_FILTER', default='true', parser=bool)
 # this should be 31536000 in prod (1 year)
 SECURE_HSTS_SECONDS = config('SECURE_HSTS_SECONDS', default='0', parser=int)


willkg:
This should have an accompanied test.

pmac:
Not sure what's to test here. This is just built into Django. What would you like to be tested @willkg?

For me this looks fine. I meant to add this before now, it just wasn't high on the list since we don't have user-uploaded content. What we really need to do is enable a strong CSP policy. I should file an issue for that actually...

freddyb:
CSP is #220 ;)

pmac:
> On a side-note, all configs default to either 'true' or 'false', which I found slightly confusing.

This is just to kind-of drive home that it's using the string parser from the [everett](https://github.com/willkg/everett) library. All parsers act on strings only, and the bool one accepts many different values for `True` or `False`.

willkg:
I was thinking that if having this header is important, then we should have at least one test that verifies the header is there. Pretty sure it's not hard to write, but I can't look into it right now.

pmac:
I feel more like it's testing configuration which is more for an integration test than for the test suite we have. I know we have no integration test suite at present, but it's more something that should be enabled in prod, but not necessarily elsewhere. We should test it in practice by doing regular looks at the [HTTP Observatory](https://observatory.mozilla.org/analyze.html?host=www.standu.ps).

pmac:
To be clear, I'm not saying we shouldn't add a test, but I also don't want to hold up the PR for it. But if you feel strongly about it then I'm not standing in the way. 

All that said, much of the actual functionality we have needs tests. I'm unfortunately quite guilty of working a bit fast-and-loose on the port to Django 😞

standup/settings.py
```diff
@@ -144,7 +144,7 @@ def path(*paths):
 USE_TZ = True
 
 # security
-
+SECURE_CONTENT_TYPE_NOSNIFF = config('SECURE_CONTENT_TYPE_NOSNIFF', default='true', parser=bool)
 SECURE_BROWSER_XSS_FILTER = config('SECURE_BROWSER_XSS_FILTER', default='true', parser=bool)
```
willkg: Out of curiosity, why would we ever want these False? Why make them configurable?

willkg:
Interesting... So I frame it differently. I frame it like, "we're testing a requirement we want met". The test doesn't really care how we implemented the requirement (could be configuration, could be a library, etc) and ensures that we continue to meet the requirement going forward.

If HTTP Observatory had an API that we could throw a cron job or something else that runs regularly at, that would sort of work, but by the time we've discovered the thing in question is gone, it's gone in prod and has been for some period of time.

I think if this were a different kind of thing, like "is Standup capitalized", I wouldn't bother testing it. I think I tend to place security requirements in a special bucket.

A test like this is sufficient:

```
class SecurityTestCase(TestCase):
    """Tests various security-related things"""
    def test_xsniff(self):
        resp = self.client.get(reverse('status.index'))
        assert resp.status_code == 200
        assert resp['x-content-type-options'] == 'nosniff'
```

I stuck it in `standup/standup/tests/test_views.py`. I haven't tested it, though. We can put the other security-related tests there, too.

I'd still rather not land this until we've pushed out standup v2. After that, we can switch to a continuous integration style and things get easier and less blocked up.

standup/settings.py
pmac: Great point. We should probably just hard-code these.

willkg:
@pmac Looks like @freddyb is doing the patches by editing in github and doesn't have a dev environment working, so if this looks good to you, land it and I'll throw a test together later.

willkg:
PR #253 adds the above test.

PR #227, opened 2016-09-20
Fix #220: Add CSP

### LINKED ISSUE #220: Implement Content-Security-Policy

pmac:


freddyb:
For a safe CSP, you should also set object-src to `'none'`.

pmac:
According to the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives#object-src) it falls back to "default-src" if not defined, which in our case is just `'self'`. Is that not good?

pmac:
@freddyb opinions on my last comment? I've looked into other implementations as well (like github's) and they don't set object-src at all, but they do have `default-src 'none'`.

pmac:
I've updated things to be better based on some new understanding, and I think this is a cleaner config as well. Let me know what you all think.

/cc @freddyb @willkg

freddyb:
Sorry for the long wait. Looks good.

pmac:
I think this is in good shape. Let's get it in an test it on stage.

PR #232, opened 2016-09-22
Fix #231 Add nixheaders extension and inline tags

### LINKED ISSUE #231: messages that start with # end up as h1
From https://github.com/mozilla/standup/issues/219#issuecomment-248748006

@a2sheppy said:

> Discovered that the results when you do https://standupstage.herokuapp.com/weekly/?month=2016-09 are pretty weird. :)

I looked into it. Those are messages from r1cky and it looks like they all start with a `#`. In Markdown, that converts to an h1, so it looks like it's doing the "right thing" formatting wise, but it's weird.

This issue covers looking at old Standup and seeing what it did with those. Maybe it dropped headers because this isn't a blog platform and that wouldn't help much?

pmac:


pmac:
Brings back the `nixheaders` extension and moves tags into the text.

standup/status/models.py
```diff
@@ -9,18 +9,22 @@
...
+from standup.mdext.nixheaders import NixHeaderExtension
 
 
-TAG_TMPL = '{0} <span class="tag tag-{1}">{2}</span>'
 BUG_RE = re.compile(r'(bug) #?(\d+)', flags=re.I)
 PULL_RE = re.compile(r'(pull|pr) #?(\d+)', flags=re.I)
-USER_RE = re.compile(r'(?<=^|(?<=[^\w\-\.]))@([\w-]+)', flags=re.I)
-TAG_RE = re.compile(r'(?:^|[^\w\\/])#([a-zA-Z][a-zA-Z0-9_\.-]*)(?:\b|$)')
+USER_RE = re.compile(r'(?<=^|(?<=[^\w\-.]))@([\w-]+)', flags=re.I)
+TAG_RE = re.compile(r'(?:^|[^\w\\/])#([a-z][a-z0-9_.-]*)(?:\b|$)', flags=re.I)
```
willkg: Making this `re.I` was a good idea. 👍

standup/status/models.py
```diff
@@ -189,11 +193,24 @@ def htmlify(self):
         # Remove icky stuff.
         formatted = bleach.clean(self.content, tags=[])
 
+        # linkify urls
+        formatted = urlize(formatted, trim_url_limit=32)
```
willkg: Moving the bits to here is a great idea. 👍

standup/status/models.py
```diff
@@ -189,11 +193,24 @@ def htmlify(self):
...
         # Linkify "bug #n" and "bug n" text.
         formatted = BUG_RE.sub(
             r'<a href="http://bugzilla.mozilla.org/show_bug.cgi?id=\2">\1 \2</a>',
             formatted)
 
+        # Wrap tags in a span for formatting
+        checked = set()
+        for tag in TAG_RE.findall(formatted):
+            if tag in checked:
+                continue
```
willkg: I don't think I understand what `checked` is doing. It seems to get set, then checked, but it doesn't seem to be modified in this for block, so won't it always be empty?

standup/status/models.py
```diff
@@ -213,17 +230,7 @@ def htmlify(self):
...
-        tags = TAG_RE.findall(self.content)
-        tags.sort()
-        if tags:
-            tags_html = ''
-            for tag in tags:
-                tags_html = TAG_TMPL.format(tags_html, tag.lower(), tag)
-            formatted = '%s <div class="tags">%s</div>' % (formatted, tags_html)
-
-        return formatted
+        return Markup(MD.convert(formatted))
```
willkg: I thought I pulled in some tests from the old standup related to formatting status posts, but you didn't edit any. So one of two things is possible:

1. Travis gets fussy with this PR.
2. We don't have these tests. If this is the case, I'll write up an issue and write some later.

standup/status/models.py
pmac: Doh! Forgot that bit. Good catch.

standup/status/models.py
pmac: Tests did fail. I fixed them I think.

willkg:
Oh noes... More bleach usage. Whoever maintains that project should get their act together.

pmac:
I think this is looking good now. How we feelin?

pmac:
I ask again because other things changed (like jinja's urlize -> bleach's linkify).

willkg:
This looks good to me. r+

PR #234, opened 2016-09-23
Rework 500 template

### LINKED ISSUE #233: 500 page errors out
Issue #230 is about an unhandled error that caused Django to go the unhandled error `->` 500 page route, but then that kicked up an error.

This issue covers fixing the `500.html` template so that it renders correctly in the event of an error.

willkg:
This nixes use of "settings" in the 500 template. For some reason,
that was causing it to bomb out. This also nixes anything that might
kick off a db query.

I think this will fix #233. If not, it'll get us closer.

Turns out it's super hard to write tests that correctly test the "unhandled exception was thrown" situation because the Django TestClient makes it impossible and DEBUG mode shows you something other than your 500 template.

If we're intent on testing this in the unit test suite, then we might have to write our own mini-test client somehow.

### DIFF
--- a/standup/status/jinja2/500.html
+++ b/standup/status/jinja2/500.html
@@ -1,9 +1,52 @@
-{% extends "base.html" %}
 {% set title = '500: Something went wrong...' %}
 
-{% block before_content %}{% endblock %}
-{% block content %}
-  <div class="grid_8 prefix_2 suffix_2">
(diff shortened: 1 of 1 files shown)


willkg:
@pmac Does this look good enough?

willkg:
@pmac contacted me from another plane of existence using his mental powers and said this is good. I will merge it now.

PR #238, opened 2016-09-23
Fix NoReverseMatch errors

pmac:
Some slugs for things in the database apparently don't match the slug regex. So when trying to display those it will rightly throw a NoReverseMatch exception.  This deals witht those by simply moving on since those are anomalies or the result of XSS testing.

### DIFF
--- a/standup/status/models.py
+++ b/standup/status/models.py
@@ -2,7 +2,7 @@ import re
 from collections import OrderedDict
 
 from django.contrib.auth.models import User
-from django.core.urlresolvers import reverse
+from django.core.urlresolvers import reverse, NoReverseMatch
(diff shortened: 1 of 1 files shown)


pmac:
Saw these [in Sentry](https://sentry.prod.mozaws.net/ELOE/standup-prod/issues/351854/). I fixed it in prod by deleting the user named `"><script>alert(1)</script>` ಠ_ಠ

pmac:
@willkg r?

willkg:
I think we should go through and ditch/fix junk data at some point. Particularly the user and team tables. I'm not sure we care about other tables.

This looks good to me. r+

PR #239, opened 2016-09-23
Add temporary "Standup v2!" announcement line

willkg:
@pmac r?


willkg:
I haven't seen what it looks like, yet, since I'm having problems getting Standup to work locally.

willkg:
^^^ That makes it a success notice so it should be better now.

willkg:
I think the spirit of @pmac contacted me through the spirit world and told me this is good enough for now.

willkg:
(But seriously, the nature of this change loses its efficacy the longer we leave it. And if we leave it over the weekend, we might as well not do it.)

PR #242, opened 2016-09-25
Fix npm install

willkg:
Previously, npm install installed into /app/node_modules, but then
that gets stomped on when copying . to /app. This reorders some things
so that doesn't happen.


willkg:
I'm not wild about this, but it does fix the issue.

@pmac Thoughts on this?

willkg:
The other way to do this is to install globally, but then we end up with paths like `/usr/local/lib/standup/node_modules/.bin/lessc`.

We could copy everything in `.bin` (they're all links) to `/usr/local/bin`, then have less screwy paths. I don't know if that's an odd thing to do or not.

I don't know enough about node to know better ways to install. Maybe there's a better way.

pmac:
This way doesn't solve the problem of node deps getting stomped on when running via `docker-compose`. When running like that, it sets your local directory as `/app` in the running container, which includes the `/app/node_modules` path. So it'd be using the node binary in the container, but your local `node_modules` dir. I think we should try installing globally, or at least to another location. I think @jgmize has dealt with this before. Perhaps he has some ideas as well.

willkg:
#243 is mo better. Closing this out for that one.

PR #243, opened 2016-09-25
Fix npm install location

pmac:
Alternate fix to #242


pmac:
@willkg this is my suggestion. Should solve both the app copy and the dev-mode running with the volume defined. Let's try these out and discuss.

willkg:
Looks good to me! r+

PR #244, opened 2016-09-25
Switch JS and CSS compressors

pmac:
Switching from the seemingly abandoned yuglify to:

* [clean-css](https://www.npmjs.com/package/clean-css)
* [uglify-js](https://www.npmjs.com/package/uglify-js)

This is what MDN is currently using and what bedrock is likely switching to shortly.


pmac:
Let's wait to merge this until after we resolve the Node installation issues with either #242 or #243, then update this for that. But we can discuss this now if anyone has objections.

willkg:
I have no preference, so it's all good to me.

PR #248, opened 2016-09-26
Add middleware to enforce a hostname of our choice

pmac:
e.g. requests to https://standups.herokuapp.com would redirect to https://www.standu.ps if we set ENFORCE_HOSTNAME=www.standu.ps.


pmac:
@willkg r?

willkg:
Sounds good to me. r+

PR #249, opened 2016-09-26
Add contribute.json file for contribution discovery

pmac:
See https://www.contributejson.org/

### DIFF
--- a/contribute.json
+++ b/contribute.json
@@ -1,0 +1,32 @@
+{
+    "name": "Standup",
+    "description": "Web app that logs daily standup updates",
+    "repository": {
+        "url": "https://github.com/mozilla/standup",
+        "license": "BSD-3-Clause",
+        "tests": "https://travis-ci.org/mozilla/standup/"
+    },
+    "participate": {
+        "home": "https://github.com/mozilla/standup",
+        "docs": "https://github.com/mozilla/standup#readme",
+        "irc": "irc://irc.mozilla.org/#standup",
+        "irc-contacts": [
+            "pmac",
+            "willkg"
+        ]
+    },
+    "bugs": {
+        "list": "https://github.com/mozilla/standup/issues",
+        "report": "https://github.com/mozilla/standup/issues/new"
+    },
+    "urls": {
+        "prod": "https://www.standu.ps",
+        "stage": "https://standupstage.herokuapp.com"
+    },
+    "keywords": [
+        "python",
+        "less-css",
+        "django",
+        "html5"
+    ]
+}
--- a/root_files/contribute.json
+++ b/root_files/contribute.json
@@ -1,0 +1,1 @@
+../contribute.json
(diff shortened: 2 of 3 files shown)


pmac:
@willkg double check me on the values in the file when you get a chance, please.

contribute.json
```diff
@@ -0,0 +1,32 @@
+{
+    "name": "Standup",
+    "description": "Web app that logs daily standup updates",
+    "repository": {
+        "url": "https://github.com/mozilla/standup",
+        "license": "BSD-3-Clause",
+        "tests": "https://travis-ci.org/mozilla/standup/"
+    },
+    "participate": {
+        "home": "https://wiki.mozilla.org/Standup",
```
willkg: So this is the Mozilla-centric page covering how we're running Standup, but it's not a page that's relevant to the Standup product which is in this repo. I'm not sure how to disentangle Standup the product vs. Standup the Mozilla service. This is a problem I have with `contribute.json` in general.

If someone else were to stand up a Standup, we'd end up with all the support requests unless they stomped on this file somehow which is tricky unless they fork the repo ... Messy.

willkg: I don't know what to do about that. Maybe ignore it for now because other people standing up a Standup is just a hypothetical at the moment.

contribute.json
```diff
@@ -0,0 +1,32 @@
...
+        "docs": "https://github.com/mozilla/standup#readme",
```
willkg: This is a tough one because I want to create a RTD site for Standup the product which will have things like API docs and such on it, but it won't have Mozilla Standup maintenance things on it.

What does "docs" entail? Is it the product docs or the project docs? The `contributejson.org` site doesn't describe what the fields mean.

willkg:
The values seem fine. They're Mozilla-centric, so it doesn't work for "standup the product" and will make it difficult to stand up other Standups. But I think we can just toss an issue in the issue tracker about that for now and if it becomes an issue, someone will comment and we can think about that then.

So, r+!

contribute.json
pmac: I thought about that too while prepping this PR... hmm... we could make the `/contribute.json` URL dynamic and the values set via settings, but that seems a bit extreme at this point as well... I'm not sure either. I'm happy to remove the wiki from the file though and just point at the repo as "home". Maybe that would be slightly better?

contribute.json
pmac: Perhaps for now "home" should be github and "docs" should be omitted until we setup RTD? I think of docs as the project docs.

willkg: That sounds fine to me.

contribute.json
willkg: I think I'm inclined to "land this now and write up an issue and not do anything with the issue until someone chimes in with a real-world problem".

pmac:
Hmmm... this is a tough one. I kinda think we do want one, but the file idea itself doesn't work well with deployable FLOSS web apps like this w/o marrying the project to one instance. I'm not sure what to do here. Maybe it would be best to go ahead and make the `/contribute.json` url content configurable via settings, but then the one in the actual repo would be wrong... 

I guess you're right and it's okay until it's not.

pmac:
Updated some values.

willkg:
Looks good to me. r+!

PR #250, opened 2016-09-26
Fix error pages not having context-processor data

pmac:
This should fix this issue. The error views included with django-jinja use normal view context instances.

### DIFF
--- a/standup/settings.py
+++ b/standup/settings.py
@@ -124,6 +124,7 @@ TEMPLATES = [
         },
     },
 ]
+DEFAULT_JINJA2_TEMPLATE_EXTENSION = '.html'
 
 WSGI_APPLICATION = 'standup.wsgi.application'
 
--- a/standup/urls.py
+++ b/standup/urls.py
@@ -1,21 +1,15 @@
-"""standup URL Configuration
-
-The `urlpatterns` list routes URLs to views. For more information please see:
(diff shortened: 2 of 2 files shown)


standup/settings.py
```diff
@@ -124,6 +124,7 @@ def path(*paths):
         },
     },
 ]
+DEFAULT_JINJA2_TEMPLATE_EXTENSION = '.html'
```
willkg: What does this do?

pmac: I found that in order to use these convenience error views you need to set this to the extension of your error templates unless you use the default: `.jinja`. See [the code](https://github.com/niwinz/django-jinja/blob/master/django_jinja/views/__init__.py#L51).

willkg: Fascinating.

willkg:
Sounds good to me!

It probably means we can re-introduce some of the stuff I nixed from the 500 page--namely use of `settings`.

pmac:
I think so too. It's a tough thing to test, but we can make sure it works locally with a view that throws an exception.

willkg:
It's tricky to do that, too. You have to set `DEBUG=False` and then deal with all the things that don't work the same way in that mode.

Way easier to test on -stage.

pmac:
Yeah. Fair point.

willkg:
Sorry, I thought I mentioned this earlier--this looks good to me. r+!

PR #251, opened 2016-09-26
Upgrade Django to 1.8.15

pmac:
Shouldn't affect standup since we don't yet have GA, but it's a good idea.


willkg:
Wow--that was super fast! r+

pmac:
I got bugs for the other sites. Figured I'd throw this one in while I was at it :) Thanks for the quick r+

PR #254, opened 2016-09-27
Individual status page 404 instead of 500

pmac:
View for an individual status was throwing an error from the paginator when the queryset was empty. This change forces the 404.

Note: We have to use the paginator even for this page because of what the macros that display statuses expect.


pmac:
We've gotten [a lot of these in Sentry](https://sentry.prod.mozaws.net/ELOE/standup-prod/issues/352152/).

willkg:
This looks good to me. r+

PR #255, opened 2016-09-27
Add /robots.txt

pmac:
Add setting to turn off disallow. Mainly so that we can
tell robots not to index the staging instance.

### DIFF
--- a/standup/status/views.py
+++ b/standup/status/views.py
@@ -261,3 +261,12 @@ def csp_violation_capture(request):
                           data=data)
 
     return HttpResponse('Captured CSP violation, thanks for reporting.')
+
+
+def robots_txt(request):
+    if settings.ROBOTS_ALLOW:
+        content = ''
+    else:
+        content = 'User-agent: *\nDisallow: /'
+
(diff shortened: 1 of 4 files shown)


standup/status/views.py
```diff
@@ -261,3 +261,12 @@ def csp_violation_capture(request):
...
     return HttpResponse('Captured CSP violation, thanks for reporting.')
+
+
+def robots_txt(request):
+    if settings.ROBOTS_ALLOW:
+        content = ''
+    else:
+        content = 'User-agent: *\nDisallow: /'
+
+    return HttpResponse(content, content_type='text/plain')
```
willkg: When we get caching going, we should totally cache this.

willkg:
This looks good to me. r+

standup/status/views.py
pmac: I could add some cache headers, so at least some clients would cache it.

willkg: We could do that... but I think we could also just create an issue for "let's cache things that make sense to cache" and push it off until then.

Either way is fine with me.

Issue #256, opened 2016-09-27
Add caching to things for which it makes sense


pmac:
Like the XML feeds, the /robots.txt, anonymous page loads, etc.

willkg:
We added a bunch of caching, then I went and removed a bunch of it just now.

We originally added caching because we were having site performance problems. However, those problems turned out to be a memory leak with the antediluvian way we were using markdown and we've since vanquished that harbinger of death.

We needed to figure out some caching things related to Auth0 support and in the process of doing that, decided to nix a lot of the caching since there's a trade-off in debugging and mysterious happenstances and we didn't think we were getting much value out of it.

So... I think we're good on this now. If we have specific caching needs, we can open up new issues and solve them there. Given that, closing this out.

PR #257, opened 2016-09-27
Ref #256: Add basic full-site cache

pmac:
* Minimal cache for full site
* 30 min (default) cache for atom feeds
* Max cache for robots.txt


pmac:
@willkg did kindof the minimum just to get us started. Not sure if the full-site middleware is even worth it, but it could help some. Just using the `LocmemCache` so far since we're only running one dyno anyway.

willkg:
Looks good to me! r+!

PR #259, opened 2016-09-29
Fix #210: Add support for New Relic

### LINKED ISSUE #210: site performance monitoring
It'd be nice to have some site performance monitoring like New Relic or Datadog or something so we can see what's going on.

It's not clear whether we *need* this, but it's a nice to have if we plan on any ongoing development/maintenance.

pmac:
Performance monitoring


pmac:
@willkg once this is in prod we just have to set some env vars on prod and it'll start reporting.

bin/run-prod.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash -ex
 
 ./bin/run-common.sh
-exec gunicorn standup.wsgi:application -b 0.0.0.0:${PORT:-8000} --log-file -
+exec newrelic-admin run-program gunicorn standup.wsgi:application -b 0.0.0.0:${PORT:-8000} --log-file -
```
willkg: What does this do if you don't have a valid new-relic account or creds or something? Does it prevent Standup from running?

pmac: In my testing it does not prevent it from running, it just runs. My plan is to make sure stage works w/o issue before adding any settings. We could then configure stage and make sure that works as well before going to prod and configuring that.

pmac: To test I did

* `make rebuild`
* `docker-compose up prod`

This runs the production config and everything worked well.

willkg:
Ok--sounds good to me! r+

PR #260, opened 2016-09-30
Attempt at more helful error page

willkg:
If there's a HELP_FAQ_URL defined, this will show a link to that on
the error page. It also shows a note for people trying to log in with
github and having difficulties.

We can't make it better than this since the problem occurs in the
login flow and we can't easily insert a handler there. Further, the
500 handler doesn't have any interesting context information, so we
can't figure it out in the template.


willkg:
Two things:

1. This is pretty meh, but I think it's the best we can do without doing "serious changes". I think it will help point people in the right direction and that's probably the most important thing at this stage.
2. I can't test this--will have to test it on stage.
3. It makes me think we should have a better way to test prod-like environment where `DEBUG=False`. Do we have something like that already?

@pmac r?

pmac:
> It makes me think we should have a better way to test prod-like environment where DEBUG=False. Do we have something like that already?

We do indeed. It's not in the Makefile, but I guess it should be. It's a target for docker-compose:

```shell
$ docker-compose up prod
```

That runs the prod command with `DEBUG=False`. You may need to manually run collectstatic.

standup/status/jinja2/500.html
```diff
@@ -23,6 +23,15 @@
       <div id="main-content" class="container_12 cf">
         <div class="grid_8 prefix_2 suffix_2">
           <div class="error-message">{{ title }}</div>
+          <div>
```
pmac: I was able to test this locally, and recommend [this markup](https://gist.github.com/pmac/d36987c7cde0f1c0dda71459e5afa8b1) for asthetics.

willkg:
^^^ That makes it look better and also removes a silly check. If `settings.DEBUG` is True, then you never see the error page anyhow. 🐼

PR #269, opened 2016-10-03
Irc nicks

willkg:
This fixes the view and API code to use a new field `Standup.irc_nick` for matching IRC users to Standup users.

This is a short-term stopgap solution. It has a few problems:

1. the API now uses IRC nicks, but it shouldn't be IRC-centric
2. the user has to create their account, then fill in the IRC nick they're using before things work
3. it continues to only supports one IRC network

Despite that, this should go out soon because it's probably causing problems especially for new users.

### DIFF
--- a/standup/status/migrations/0006_add_irc_nick.py
+++ b/standup/status/migrations/0006_add_irc_nick.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('status', '0005_add_team_user_id'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='standupuser',
+            name='irc_nick',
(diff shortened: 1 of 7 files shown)


willkg:
@pmac ^^^ Does this look ok?

standup/status/migrations/0006_add_irc_nick.py
```diff
@@ -0,0 +1,30 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='standupuser',
+            name='irc_nick',
+            field=models.CharField(help_text='IRC nick for this particular user',
+                                   max_length=100, null=True, unique=True, blank=True),
+        ),
+        migrations.AlterField(
```
pmac: Not sure why `github_handle` and `slug` are being modified in the migration. They don't seem to have been touched in models.py.

willkg: I suspect we adjusted them in a previous change, but never rolled new migrations. I think it'd take like 5-10 minutes to look into, but it seemed pretty innocuous so I didn't.

willkg:
I'm merging and will test on -stage.

PR #273, opened 2016-10-04
Db opts

willkg:
Add .select_related() calls to reduce DB back-and-forth
    
This doesn't completely eliminate the db back-and-forth. The home page still does an auth_user and a status DB query for every status it shows on the page. The auth_user one is to get the email address to
show the gravatar. The status one is to get any replies.
    
Despite that, this should improve performance.

After this, we should look at moving the email field or caching it in `StandupUser`. After that, we could rethink reply_to or investigate optimizing htmlify.

r?

### DIFF
--- a/standup/settings.py
+++ b/standup/settings.py
@@ -351,6 +351,23 @@ RAVEN_CONFIG = {
     'release': config('GIT_SHA', raise_error=False),
 }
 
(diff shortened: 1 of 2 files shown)


willkg:
@pmac Does this look ok?

PR #289, opened 2016-12-01
Tweak cache settings for better memory usage

willkg:
We're running out of memory in prod constantly. My current theory is
that our cache settings are such that we're using an increasing amount
of cache until we pass the threshold for memory allowance.

Assuming that's correct, limiting the caching by number of entries and
timeout should make that better.

This implements that.

### DIFF
--- a/standup/settings.py
+++ b/standup/settings.py
@@ -147,6 +147,11 @@ CACHES = {
                       parser=django_cache_url.parse,
                       default='locmem:default')
 }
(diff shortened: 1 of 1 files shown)


willkg:
Relevant docs: https://docs.djangoproject.com/en/1.8/topics/cache/#cache-arguments

willkg:
This is intentionally aggressive. I want to see if this affects things. If it does, then we have proved that cache is part/all of the issue and we can figure out what we want to do from there.

PR #292, opened 2016-12-02
Update requirements

willkg:
This updates a bunch of requirements. Minor changes, so bug fixes and
the like.

Figured I'd do this before I continued on memory stuff in case one of these things had a fixed memory leak.


willkg:
Self-r+!

PR #293, opened 2016-12-02
Fix markdown call

willkg:
We use the same Markdown instance for all convert calls. Per the
Markdown docs, we need to call .reset() between convert calls so as to
reset the Markdown converter state and thus avoid performance
degredation over time.

https://pythonhosted.org/Markdown/reference.html#Markdown

This does that.

This also has a HUGE effect on memory and response time performance.


willkg:
It's a super minor change. self-r+.

pmac:
ZOMG! That was it!!! Good work @willkg! 💯 🎉

PR #294, opened 2016-12-02
Fix statistics page rendering

willkg:
Right now, it's kicking up an error that suggests it's missing context. It worked fine locally and in the tests, so I'm puzzled.

Regardless, this tweaks the code a bit to see if it's better.


willkg:
I think this is ok. Going to land it. self-r+.

Issue #280, opened 2016-10-29
remove persona


willkg:
We've had Github auth for a month. It's time to remove Persona.

willkg:
We removed Persona and replaced it with Auth0. Closing this out.

PR #296, opened 2016-12-07
Remove Persona

willkg:
This removes Persona and django-browserid and all that from the
codebase. It leaves some goofy pages (e.g. the login page), but I think
that's an ok stop-gap for now as this is only the first step in
rewriting the auth flow to use auth0.

r?


willkg:
@pmac Does this look ok? Did I miss anything?

willkg:
Fixes #280.

willkg:
I was thinking that we could break everything and then from the ashes build a new auth flow.

PR #299, opened 2016-12-16
Drop github auth for auth0

### LINKED ISSUE #295: switch to auth0
@pmac and I spent some quality time with top people and learned much about auth0 and we want to switch to it. It solves the following problems:

1. should fix our auth flow which seems broken for certain groups of people
2. allows for passwordless login so users don't have to have a github account
3. radically simplifies our auth flow code

This issue covers switching to the new system.

willkg:
This reworks the authentication flow to use Auth0 and only Auth0. It
also cleans up some code.

Doing auth0 is kind of hard, so there aren't any tests for the auth
flows. Those have to be tested by hand for now.

Many thanks to PeterBe. Looking at how Air Mozilla implemented Auth0 was
super helpful.

Fixes #295


willkg:
This changes the auth flow. I looked at the data in -prod and we've got user accounts in "interesting states". Because we've switched auth systems a few times and each one has been more exciting and splendiferous than the previous one and they have all had different assumptions and requirements, many user accounts are in "interesting states". This attempts to fix some of those states, but we'll definitely have to fix a bunch by hand.

We have around 40 active users right now. I'm pretty ok with fixing accounts by hand for a bit.

willkg:
So I don't forget, when we land this and push it out, we'll want to update the wiki page, too.

standup/settings.py
```diff
@@ -72,8 +70,8 @@ def path(*paths):
     'django.contrib.auth.middleware.SessionAuthenticationMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
-    'social.apps.django_app.middleware.SocialAuthExceptionMiddleware',
     'django.middleware.cache.FetchFromCacheMiddleware',
+    'standup.status.middleware.NewUserProfileMiddleware',
```
willkg: Moved this to the bottom so that it picks up the user from the session.

standup/settings.py
```diff
@@ -188,6 +185,21 @@ def path(*paths):
...
 HELP_FAQ_URL = config('HELP_FAQ_URL', raise_error=False)
 
+# auth
+AUTHENTICATION_BACKENDS = (
+    'django.contrib.auth.backends.ModelBackend',
+)
+LOGIN_URL = '/'
+LOGOUT_URL = '/logout/'
+LOGIN_REDIRECT_URL = '/'
+LOGIN_REDIRECT_URL_FAILURE = '/'
```
willkg: I'm not sure we need these four anymore. I'll check on Monday.

standup/settings.py
```diff
@@ -301,6 +284,15 @@ def path(*paths):
             ),
             'output_filename': 'js/common.min.js',
         },
+        'signin': {
+            'source_filenames': (
+                'js/vendor/jquery-3.1.0.js',
+                'js/vendor/lock-passwordless-2.2.min.js',
```
willkg: lock-passwordless-2.2.min.js is biggish, so I figured it made sense to keep a separate signin page so we didn't have to download it with other pages.

Yes, we have caching and all that, but seemed like a better idea to just keep other pages minimal.

standup/status/auth.py
```diff
@@ -1,19 +0,0 @@
...
-                user=user,
-                name=details['fullname'],
-                slug=response['login'],
-                github_handle=response['login'],
-            )
-        else:
-            if profile.github_handle != response['login']:
-                profile.github_handle = response['login']
-                profile.name = details['fullname']
-                profile.save()
```
willkg: This was for the github backend which we don't use anymore, so I nixed the file.

standup/status/jinja2/base.html
```diff
@@ -72,9 +73,10 @@
                     {% if user.is_staff -%}
                       <li><a href="{{ url('admin:index') }}">Admin</a></li>
                     {%- endif %}
+                    <li><a href="{{ url('users.logout') }}">Logout</a></li>
```
willkg: Previously, there was no way to log out, so I added that. This really should be a POST action with a csrf, though.

standup/status/jinja2/users/login.html
```diff
@@ -1,20 +1,56 @@
...
+
+{% block more_styles %}
+  {% stylesheet 'user' %}
+{% endblock %}
+
+{% block title %}{{ title }} | {{ super() }}{% endblock %}
+{% block more_meta %}
+  <meta name="auth0-client-id" content="{{ settings.AUTH0_CLIENT_ID }}">
+  <meta name="auth0-domain" content="{{ settings.AUTH0_DOMAIN }}">
+  <meta name="auth0-callback-url" content="{{ settings.AUTH0_CALLBACK_URL }}">
```
willkg: The javascript pulls these from the page and uses them.

standup/status/jinja2/users/login.html
```diff
@@ -1,20 +1,56 @@
...
+        <form>
+          <button class="btn login-button" type="submit"><span>Sign in using Auth0</span></button>
+        </form>
+      </div>
+    {% endif %}
+  </div>
+{% endblock %}
+
+{% block javascript %}
+  {% javascript 'signin' %}
```
willkg: This page is ugly. After this lands, I'm going to bug @rehandalal and see if he can make it less ugly.

standup/status/views.py
```diff
@@ -160,11 +164,185 @@ class LoginView(TemplateView):
...
+        except (ConnectTimeout, ReadTimeout):
+            messages.error(
+                request,
+                'Unable to authenticate with Auth0 at this time. Please wait a bit '
+                'and try again.'
+            )
+            return HttpResponseRedirect(reverse('users.loginform'))
+
+        # Get or create User instance
+        user = get_or_create_user(user_info['email'], user_info.get('name'))
```
willkg: TODO: I need to verify what "name" is here. Is it a full name? (e.g. Will Kahn-Greene) Or a nick name? (e.g. willkg)

standup/status/views.py
```diff
@@ -160,11 +164,185 @@ class LoginView(TemplateView):
...
+
+        # If inactive, add message and redirect to login page
+        if not user.is_active:
+            messages.error(
+                request,
+                'This user account is inactive.'
```
willkg: We don't currently have inactive users in this sense. We'd probably use this to terminate accounts. If that's the case, then we'd probably want a different message here.

Regardless, this is probably fine for now?

willkg:
^^^ That covered all the things I think I missed.

@pmac Got some time to look this over? We'll do testing on -stage of flows, so mostly this is just to catch things before we get there.

standup/status/views.py
```diff
@@ -160,11 +164,185 @@ class LoginView(TemplateView):
...
         return super().get(request, *args, **kwargs)
 
 
+class LogoutView(View):
+    """Logs a user out"""
+    def post(self, request, *args, **kwargs):
+        if not request.user.is_authenticated():
+            messages.info(request, 'You are not logged in.')
+        else:
+            # FIXME(willkg): Should require a csrf.
```
pmac: Should automatically check CSRF if the CSRF middleware is enabled, which it is.

standup/status/views.py
```diff
@@ -160,11 +164,185 @@ class LoginView(TemplateView):
...
+
+        # Make sure they have a profile
+        get_or_create_profile(user)
+
+        # Log the user in
+        # FIXME(willkg): This is sort of a lie--should we have our own backend?
+        user.backend = 'django.contrib.auth.backends.ModelBackend'
+        login(request, user)
+
+        # FIXME(willkg): Should we send them to the front page or their user page?
```
pmac: front page seems fine

standup/status/views.py
```diff
@@ -160,11 +164,185 @@ class LoginView(TemplateView):
...
+            logout(request)
+            messages.info(request, 'You have been logged out.')
+        return HttpResponseRedirect('/')
+
+
+def username_generator(base):
+    """Generates usernames using a base plus some count"""
+    yield base
+
+    count = 2
```
willkg: Oops--both good ideas. I'll fix this now.

willkg:
^^^ That includes changes base on the review. I'm going to land this and test it out on stage.

PR #301, opened 2016-12-20
Change Auth0 so it's not required

willkg:
This switches the code around so that Auth0 is not required. If the
configuration is not available, then signin is disabled.


willkg:
@pmac Does this look ok? Pretty sure it's "good enough" for now. We can adjust the signin disabled stuff as we need to.

PR #302, opened 2016-12-20
Implement "make createsuperuser"

### LINKED ISSUE #268: fix auth to allow for easier development
Right now if you want to do local development, you have to use Persona auth because that doesn't require tokens. If you want to use GitHub auth, you have to set up tokens and blah blah blah.

When Persona goes away, we should have some kind of fake-GitHub auth working so someone can spin up a docker environment, create an account and be good to go.

willkg:
This has some instructions and creates a super user. I also added a note to the
README about how to log in for local development.

Fixes #268


.editorconfig
```diff
@@ -9,6 +9,9 @@ end_of_line = lf
 charset = utf-8
 indent_style = space
 
+[Makefile]
+indent_style = tab
```
pmac: brilliant! I'm always having this problem.

`/me adds this to his default .editorconfig`

🎉

Makefile
```diff
@@ -82,4 +82,10 @@ docs: .docker-build
 	${DOCKERCOMPOSE} run web $(MAKE) -C docs/ clean
 	${DOCKERCOMPOSE} run web $(MAKE) -C docs/ html
 
+createsuperuser: .docker-build
+	@echo "Create a superuser account for Standup local dev."
+	${DOCKERCOMPOSE} run web ./manage.py createsuperuser
```
pmac: If they do this first it's possible they'll not have run `manage.py migrate` yet right? Perhaps we should have a `make migrate` that depends on `.docker-build` and have this depend on `migrate`?

willkg: Oh, hrm... I'll have to think about this.

willkg:
I fixed `.editorconfig` in a separate commit.

This PR covers setting up a local dev environment with and without an auth0 account. I created #354 to cover that.

The rest of this should go away, so I'm going to close it out.

PR #304, opened 2016-12-22
Remove Standup v2 announcement notice

willkg:
This notice is temporary. At this point, we're pretty far into Standup v2 that
we don't need it anymore.


willkg:
This is a small change. self-r+ and all that.

PR #306, opened 2016-12-22
Tables fix

### LINKED ISSUE #305: remove social auth tables
When removing python-social-auth, I didn't remove the tables that it uses. So we're in this situation where there are foreign keys from those tables, but Django doesn't know about them so shenanigans.

We need to remove those tables because they break the Django admin.

willkg:
Fixes #305.


willkg:
Notes: Before landing this and before pushing this to -prod, we should back up the relevant databases.

willkg:
This is the PR that keeps on giving.

willkg:
This is all messed up. I tried to switch Travis from running the tests in postgres to running them in sqlite3 in #307, but I missed the export in `bin/run-tests.sh` which sets it to sqlite3.

Fixing that unearths a bunch of new problems. Running tests in test vs. test-image. Making sure to create the db before running the tests. Etc.

I'm going to manually drop those tables in -prod now and we can figure out what we want to do here in January.

willkg:
Issue #310 is still a problem, but we should solve that in that issue. Everything else that was being worked on in here is either not a problem anymore or something something. Closing this out.

PR #307, opened 2016-12-22
Switch Travis to use postgres

willkg:


willkg:
I see `DATABASE_URL` in the travis output, so I think this ran the tests with the changed `.travis.yml` file. I'm going to merge this now and wait for it to go to -stage and then make sure the changes here didn't hose -stage. I don't think they will, but I don't have a good mental model for some of the interactions between the `.travis.yml` file and what the resulting docker container looks like.

pmac:
Could also have spun up a postgres docker image and used that. This works fine though I think.

willkg:
It actually doesn't work at all. We have the `DATABASE_URL` hardcoded to be sqlite, so then it continues to use sqlite.

We need to refigure this. :(

Issue #308, opened 2016-12-22
Enhance auth0 login support


gdestuynder:
Hi - I discussed this with :willkg on IRC prior to opening this issue.

Regarding the new Auth0 login for standup, I'm proposing to:

1) Use what we call our "hosted lock" which is hosted on auth0's properties on our behalf. The reason for this are:
a) it makes the experience more consitent for users (any update there is reflected for all sites (as in relying parties/RPs) )
b) ensure password managers save the passwords for a single domain instead of every RP's domains, though this is specific to if you enable LDAP in the future

2) Not use Auth0 branding on the login button (currently it says "Login with Auth0") but instead something like "Login to standup". We have sample buttons as well if you want to re-use them.

3) Ensure you're able to refresh user sessions (i.e. if a user gets disabled because his account is compromised, that the account can no longer login to standu.ps within a certain amount of time).

Here's docs and examples for all this:

- Hosted lock demo: http://testrp.security.allizom.org/ (code: https://github.com/mozilla-iam/testrp.security.allizom.org/)
- Login buttons: https://github.com/mozilla-iam/testrp.security.allizom.org/tree/master/sign-in_buttons (though you can use your own if you like as long as it's not branded)
- Session handling spec: https://wiki.mozilla.org/Security/Guidelines/OpenID_Connect#Session_handling (and example: https://github.com/mozilla/airmozilla/blob/9fbb04c7753376e539c744b6634b06a2d2d3ecfb/airmozilla/authentication/auth0.py  https://github.com/mozilla/airmozilla/blob/9fbb04c7753376e539c744b6634b06a2d2d3ecfb/airmozilla/authentication/middleware.py)

Notes: we plan to enhance our docs to make these more clear as well, when you request integration with the Mozilla SSO with OIDC and Auth0.

willkg:
Grabbing this. I'll probably do it in a couple of passes.

willkg:
I'll do it with one pass. I fixed 1 and 2 already in a local branch. I'll work on item 3. I can lift that from Air Mozilla.

Also, this begs the question "should we throw this in a lib" and the answer is probably "yeah", so I'll think about that, too.

willkg:
The new auth0 flow works on -stage.

I still need to implement item 3.

gdestuynder:
It would be really cool in a lib. I would use it  and put it on testrp.security.allizom.org as a demo for others.
Note, theres some libs around though they don't do the session handling for example
[...]

PR #309, opened 2016-12-23
Change language

willkg:
This fixes items 1 and 2 in #308.

I need to see what this does on -stage before I can fix item 3.


willkg:
I'm going to land this so I can see what happens on -stage.

standup/status/views.py
```diff
@@ -159,21 +156,20 @@ def form_invalid(self, form):
...
     def get_context_data(self, **kwargs):
         ctx = super().get_context_data(**kwargs)
         ctx['auth0_configured'] = is_auth0_configured()
+
+        ctx['auth0_login_url'] = settings.AUTH0_LOGIN_URL.format(
+            AUTH0_DOMAIN=settings.AUTH0_DOMAIN,
+            AUTH0_CLIENT_ID=settings.AUTH0_CLIENT_ID,
+            AUTH0_CALLBACK_URL=settings.AUTH0_CALLBACK_URL,
+            # FIXME(willkg): This should be a token that ties both ends.
+            STATE='foo',
```
pmac: Should we fix this before prod? Is it just some string that should be unique per instance? A setting?

willkg: Yes--it should. I needed to see what this did on -stage because I can't use the mozilla.auth0.whatever thing on my local dev instance. That's why I merged this.

Outstanding things to do before we push to -prod:

1. do the `id_token` work which involves renewing the token
2. go through and fully test all the authentication options
3. plus figure out this STATE thing

Issue #279, opened 2016-10-26
standups seems to assume my github name is the same as my IRC name


captainbrosset:
I just signed in to standups, choosing the github login thing.
My github name is @captainbrosset, but my IRC nick is pbro.
When I post an update from IRC, the bot tells me https://www.standu.ps/user/pbro/ which leads to a 404, because my "page" on that website really is https://www.standu.ps/user/captainbrosset/

So it looks like the bot defaults to using the IRC nick in the URLs it posts to IRC, even if the URL doesn't really exist.

Sorry if this issue has already been filed, I couldn't find it.

willkg:
Sorry about that--I totally thought I replied to this issue, but I didn't.

The idea of "identity" suddenly got tricky in the standup universe. We have a bunch of different identities across different systems that all map to a Standup identity. The problem is that this wasn't quite right in the old Standup (in my opinion) which conflated irc nick with standup id. We fixed some of the issues with v2, but we didn't really fix it. Issue #212 covered some of the problems.

I think what we need to do here is fix the API so that it returns information that the standup-irc bot can use to generate the message rather than have standup-irc bot guess it (poorly).

PR #313, opened 2016-12-26
Fix #312: Update SLUG_RE to accept backticks

### LINKED ISSUE #312: User page is 404 when username contains a backtick (`)
My IRC nickname is *NiKo`* (notice the backtick at the end) and it seems the system doesn't accept this char in the urls, so my page is 404 : https://www.standu.ps/user/NiKo%60/

Also and as a probable side effect, my entries are not linked to my personal profile page but to [the team's one](https://www.standu.ps/project/storage-team/) (though maybe that's a setting? idk).

Feels like some regex needs to be updated anyway. (Edit: #313)

n1k0:
Refs #312. Sorry I didn't test the patch, I've just edited this file right from github. Feel free to close or take over if that's not the right way to properly address the issue :)

### DIFF
--- a/standup/status/migrations/0005_add_team_user_id.py
+++ b/standup/status/migrations/0005_add_team_user_id.py
@@ -21,11 +21,6 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.AddField(
-            model_name='teamuser',
-            name='id',
-            field=models.AutoField(serialize=False, auto_created=True, verbose_name='ID')
-        ),
         migrations.RunPython(fill_teamuser_ids, reverse_code=lambda x, y: None),
         migrations.AlterField(
             model_name='teamuser',
--- a/standup/status/urls.py
+++ b/standup/status/urls.py
@@ -5,7 +5,7 @@ from django.views.decorators.cache import cache_page
 from . import views
 
 
-SLUG_RE = r'(?P<slug>[-a-zA-Z0-9_@`]+)'
+SLUG_RE = r'(?P<slug>[-a-zA-Z0-9_@`\[\]\{\}\^]+)'
 cache_feed = cache_page(settings.CACHE_FEEDS_SECONDS)
 
 urlpatterns = [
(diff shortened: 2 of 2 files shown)


glasserc:
If this is meant to match IRC nicks, it seems like https://tools.ietf.org/html/rfc1459#section-2.3.1 says that nicks can include "special" characters including '-', '[', ']', '\', '`', '^', '{', and '}'.

n1k0:
I couldn't manage including the backslash character as Django seems to intercept it early and rewrite it to a regular slash.

standup/status/migrations/0005_add_team_user_id.py
```diff
@@ -21,11 +21,6 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.AddField(
-            model_name='teamuser',
-            name='id',
-            field=models.AutoField(serialize=False, auto_created=True, verbose_name='ID')
-        ),
```
n1k0: See #314

n1k0:
Hmm this patch may just be plain wrong after all. After reading the codebase, it appears we should always rely on Django's `User` model `username` property to resolve user urls. So, a few thoughts:

- Maybe what needs to be updated first is the [standup_irc](https://github.com/mozilla/standup-irc) bot, which is generating erroneous urls;
- But as it consumes the API, maybe what should be updated as well is its views exposing user information, so the bot has access to the appropriate data to forge urls;
- The Auth0 integration seems to [automatically generate usernames](https://github.com/mozilla/standup/blob/ddfb37bd07e5833986fe2622ac59badc92d6df50/standup/auth0/utils.py#L21-L30), maybe a good idea would be to ask users to provide one they're happy with instead? So we would also have our unique slugified identifier to build proper urls.

I'm gonna move to other work I have to do, hope this investigation was somewhat helpful.

willkg:
You're seeing issue #279. After we fix that, the problem you're hitting should go away. I'm not sure we should additionally adjust slugs, but we can create a new issue for that and discuss it there if you like.

For now, I'm going to close this out.

PR #316, opened 2017-01-03
Redo auth0 implementation

willkg:
This redoes the entire auth0 implementation as a separate app with a possibility
of converting it to a standalone library.

This implements the id_token renewing middleware as well as a model, but there
are no tests for it, no migration and it's not hooked up. I'll get there after I
see what things look like.


standup/auth0/middleware.py
```diff
@@ -0,0 +1,121 @@
+from datetime import datetime, timedelta
```
willkg: None of this is tested or hooked up to anything in this PR. So the code in this file is just hanging out. Feel free to skip it.

standup/auth0/models.py
```diff
@@ -0,0 +1,13 @@
+from django.conf import settings
```
willkg: There's no migration for this, yet. It's not hooked up to anything. Feel free to skip it.

standup/auth0/pipeline.py
```diff
@@ -0,0 +1,111 @@
+import logging
```
willkg: This is the new stuff.

I broke up the callback view into a set of units which someone can assemble into an authentication pipeline depending on their needs. It's very very very loosely based on the one in python-social-auth.

Which begs the question "why not integrate this with python-social-auth?" I'm not sure we can because the bulk of the work for auth0 is done in the callback view and not in an auth backend.

standup/auth0/views.py
```diff
@@ -59,7 +58,7 @@ def get(self, request):
                     'Unable to authenticate with Auth0 at this time. Please refresh to '
                     'try again.'
                 )
-                return HttpResponseRedirect(reverse('users.loginform'))
+                return HttpResponseRedirect(reverse(settings.AUTH0_SIGNIN_VIEW))
```
willkg: There are a few changes like this one which generalize this code so we can librarificate it more easily later on.

willkg:
@pmac Can you skim this and let me know if there are terrible things here? Theoretically, it should be a rewrite of what we've got on -stage now.

From here, we need to:

1. implement the `id_token` stuff
2. fix STATE
3. go through and test the different auth0 providers we have on -stage to make sure they all work

Why land this now? I need to see what we're getting back from the Mozilla auth0 thing in the `user_info` thing for mozilla.com and non-mozilla.com addresses. That way I can figure out how to do the next round of things.

pmac:
Looks good to me. Let's land it and see what's really going on. Thanks for all of this work!

willkg:
Thank you!

PR #319, opened 2017-01-06
Quote links, move things around, tweaks

willkg:
This moves some Auth0 stuff around so more of it is in the auth0 app. Further,
it fixes auth0 url generation so that the values are quoted and it fixes a
typo (redirect_url -> redirect_uri).


willkg:
This is pretty benign. Landing this to test it out on -stage.

PR #322, opened 2017-01-09
Fix #321: Do bleach linkify after markdown->html

### LINKED ISSUE #321: markdown links with hashes in link text get corrupted
In writing my most recent status update, I wrote:

```
[#3510](https://github.com/rust-lang/cargo/pull/3510)
```

which I expected to get linkified like: [#3510](https://github.com/rust-lang/cargo/pull/3510)

Instead, it got linkified as shown in [my status update](https://www.standu.ps/status/40927/) (the "`cargo` test fixes" bullet), which...I'm not quite sure if there's a better way to describe it than looking at the result.

I assume this is a result of the `#` character in the link text, but I'm not really sure...

pmac:
Turns out bleach is meant to understand HTML, not Markdown.  Also add some tests for the rest of the htmlify method.

### DIFF
--- a/standup/status/models.py
+++ b/standup/status/models.py
@@ -20,9 +20,9 @@ from standup.status.utils import (
 from standup.mdext.nixheaders import NixHeaderExtension
 
 
-BUG_RE = re.compile(r'(bug) #?(\d+)', flags=re.I)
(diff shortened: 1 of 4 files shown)


willkg:
Looks good to me. Yay for tests!

PR #325, opened 2017-01-10
[WIP] Add id_token handling; fixes; tests

willkg:
* Adds handling and renewal for id_tokens for domains that require it.
* Adds tests.
* Fixes a bunch of bugs in the code.
* Adds migration for the IdToken model.


willkg:
@pmac This is a work in progress. There are still a few things to do and a couple of issues we need to figure out.

Things to do:

1. write tests for the middleware and callback view and probably some other things
2. centralize the required settings and defaults into a convenient place so we don't have to do what we're doing now with settings
3. write some docs for this
4. go through the code and clean up comments and review

Things to figure out for Standup:

1. this requires a shared cache for all request handling processes--we can't just use dummy: cache since each process has a different one; maybe we move it to sessions?
2. Mozilla uses language like "Mozilla LDAP", but one of our error messages is like "Oauth2 provider". How do we make those messages general, but useful for our users?

requirements.txt
```diff
@@ -1,3 +1,19 @@
+# database
```
willkg: I moved things in this file so it's easier to do `hashin xyz requirements.txt` and it's more likely that the item falls into the right section.

requirements.txt
```diff
@@ -67,3 +67,23 @@ django-cache-url==1.3.1 \
...
+    --hash=sha256:d0509648b2074bf5a66c9d6ad3ee6d3711cc035fe38756cd02df93af2187f31b \
+    --hash=sha256:953be622e88323c6f43fad61ffd05bebe73b9fd9863a46d68b052d2aa7d71ce2 \
+    --hash=sha256:b1e260204d712059cb1f60e86f1f28b72570380b74d1cd550aac1903169d9f74 \
+    --hash=sha256:18d2f00229c8aa364a29d112d6fe7d7e988d6c0917578ac2907532698154836a \
+    --hash=sha256:9ed51dc5cffa6366aad2e1e7d7ce65954b32092dc0585a591726da14c418adbd \
+    --hash=sha256:0dc3afa149746b4e7551471ff868135c92d7b68b8ac37225388fac29127d836b \
+    --hash=sha256:ddff7c3745a75cbd98b2d148eebbd739d7d18dfb537a857b02774cc09a333129 \
+    --hash=sha256:e633def17eedd2d202437daeb8fda6f379d51a34de8b809800e2737d3e7363f9 \
+    --hash=sha256:370f2b68a93fb533b537dfacc13300928f794e7dbdf7c72769d80432ea71631a \
+    --hash=sha256:7e6f55c72388afa83c38fda0d5dab710f364476b661d8b5441ab79c1d402e9be
```
willkg: simplejson is the only new thing in this file. requests prefers to use it if it's available and now our auth0 code looks for an exception simplejson generates.

standup/auth0/middleware.py
```diff
@@ -44,78 +45,84 @@ def renew_id_token(id_token):
...
-                # If there is no IdToken, then this isn't a mozilla.com address and we're fine.
+                # If there is no IdToken, check if this domain requires them and if so, redirect and
+                # if not, we're fine.
+                if request.user.email:
+                    domain = request.user.email.lower().split('@', 1)[1]
+                    if domain in settings.AUTH0_ID_TOKEN_DOMAINS:
+                        messages.error(
+                            request,
+                            'You can\'t log in with that email address using the provider you '
+                            'used. Please log in with the Oauth2 provider.'
```
willkg: This message doesn't really work for Mozilla because it doesn't match the auth0 interface users see, so we'll need to figure out what to do about that.

standup/auth0/middleware.py
```diff
@@ -44,78 +45,84 @@ def renew_id_token(id_token):
...
+                request.user.is_active and
+                request.path not in self.exception_paths
         ):
-            # Look up expiration in session and see if the id_token needs to be renewed.
-            id_token_expiration = request.session.get('id_token_expiration', None)
-            if id_token_expiration and id_token_expiration < datetime.utcnow():
+            cache_key = 'auth0:renew_id_token:%s' % request.user.id
+
+            # Look up expiration in cache to see if id_token needs to be renewed.
+            if cache.get(cache_key):
```
willkg: Since we're using multiple processes and dummy cache, this probably won't work for us. However, if we don't stash it here, I'm not sure where else to put it other than in the db. If we use the db, that means we're doing another db hit for every request. Though maybe that's fine for sites that don't have memcache.

We don't want it in the session or in a cookie because it dictates when we try to renew this token and thus it should never get to the client.

standup/auth0/middleware.py
```diff
@@ -44,78 +45,84 @@ def renew_id_token(id_token):
...
-                    )
-                    return HttpResponseRedirect(reverse(settings.AUTH0_SIGNIN_VIEW))
+            try:
+                new_id_token = renew_id_token(token.id_token)
+            except (ConnectTimeout, ReadTimeout):
+                messages.error(
+                    request,
+                    'Unable to validate your authentication with Auth0. '
+                    'This can happen when there is temporary network '
+                    'problem. Please sign in again.'
```
willkg: Oops--this needs a `fail_silently` otherwise it flips in tests.

standup/auth0/middleware.py
```diff
@@ -44,78 +45,84 @@ def renew_id_token(id_token):
...
+                )
+                # Log the user out because their id_token didn't renew and send them to
+                # home page.
+                logout(request)
+                return HttpResponseRedirect(reverse(settings.AUTH0_SIGNIN_VIEW))
+
+            if new_id_token:
+                # Save new token and re-up it in cache
+                token.id_token = new_id_token
+                token.exp = datetime.utcnow() + timedelta(seconds=settings.AUTH0_ID_TOKEN_EXPIRY)
```
willkg: Oops--I changed the name of this.

standup/auth0/middleware.py
```diff
@@ -44,78 +45,84 @@ def renew_id_token(id_token):
...
+
+            else:
+                # If we don't have a new id_token, then it's not valid anymore. We log the user
+                # out and send them to the home page.
+                logout(request)
+                messages.error(
+                    request,
+                    'Unable to validate your authentication with Auth0. '
+                    'This is most likely due to an expired authentication '
+                    'session. You have to sign in again.'
```
willkg: This needs a `fail_silently`, too.

standup/auth0/middleware.py
willkg: This also needs a `fail_silently`.

standup/auth0/pipeline.py
```diff
@@ -33,13 +36,14 @@ def run_pipeline(pipeline, **kwargs):
     return kwargs
 
 
-def log_user_info(user_info, **kwargs):
-    """Logs user_info
+def log_info(user_info, token_info, **kwargs):
```
willkg: I should generalize this to just print everything in `kwargs`. Then you can use it to debug the pipeline.

standup/auth0/pipeline.py
```diff
@@ -109,3 +125,52 @@ def login_user(request, user, **kwargs):
...
+            messages.error(
+                request,
+                'You can\'t log in with that email address using the provider you '
+                'used. Please log in with the Oauth2 provider.',
+                fail_silently=True
+            )
+            return HttpResponseRedirect(reverse(settings.AUTH0_SIGNIN_VIEW))
+
+
+def mozilla_auth0_pipeline(**kwargs):
```
willkg: This makes it much easier for Mozilla sites to have the right pipeline and makes it easier for us to update that pipeline when we do releases.

standup/auth0/tests/test_pipeline.py
```diff
@@ -42,16 +44,25 @@ def test_user_exists(self):
             email='foo@example.com'
         )
 
-        ret = pipeline.get_user_by_username({'nickname': 'foo'})
+        ret = pipeline.get_user_by_username(
+            user_info={'nickname': 'foo'},
+            junk=123
```
willkg: The `junk` part verifies that the pipeline function has a `**kwargs` catchall.

standup/auth0/tests/test_pipeline.py
```diff
@@ -117,5 +150,103 @@ def test_active(self):
...
+        request = self.factory.get('/')
+        ret = pipeline.reject_unverified_email(
+            request=request,
+            user_info={'email_verified': False, 'email': 'foo@example.com'},
+            junk=123
+        )
+        assert isinstance(ret, HttpResponseRedirect)
+
+
+# FIXME(willkg): RequireIdTokenTestCase
```
willkg: Oops--I can nix this.

standup/auth0/middleware.py
pmac: Looks to me like our default cache is locmem, which should work if they hit the same instance, but isn't great, you're right. We could definitely add memcache if it'd help. With it coded like this we can just configure memcache later and it'll improve things.

setup.cfg
```diff
@@ -2,11 +2,8 @@
 addopts = -rsxX --reuse-db --tb=native
 norecursedirs = .git docs bin staticfiles
 testpaths = standup
-python_files=test*.py
-DJANGO_SETTINGS_MODULE=standup.settings
-SECRET_KEY=foo
```
willkg: These were ignored because we're not using a pytest plugin that puts these in the environment. So I nixed them.

standup/auth0/middleware.py
```diff
@@ -44,78 +45,93 @@ def renew_id_token(id_token):
...
                 token = IdToken.objects.get(user=request.user)
+
             except IdToken.DoesNotExist:
-                # If there is no IdToken, then this isn't a mozilla.com address and we're fine.
+                # If there is no IdToken then something is weird because they should have one. So
+                # log them out and tell them to sign in
+                messages.error(
+                    request,
+                    'You can\'t log in with that email address using the provider you '
+                    'used. Please log in with the Oauth2 provider.',
```
willkg: We need to figure out what to do with this and other messages in this file.

standup/auth0/models.py
```diff
@@ -3,11 +3,11 @@
...
 
-    These tokens need to be renewed on a periodic basis and if they fail, they cause the Django
-    session to expire.
+    These tokens need to be renewed on a periodic basis and if they fail, they force Django logout.
 
     """
     user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
-    id_token = models.TextField(max_length=100, null=True)
+    id_token = models.TextField(help_text='The id token for this user.')
+    expire = models.DateTimeField()
```
willkg: I should add a `help_text`. Otherwise it causes a new migration when we add one.

willkg:
Outstanding things to do:

1. write tests for callback view
2. review the code and clean it up
3. fix error messages
4. figure out what to do about matching users and nicknames -> usernames
5. make `github_handle` not have to be unique and allow nulls (later we should probably nix it)

standup/auth0/settings.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def get(self, key):
+        opt = self._option_map[key]
+        val = getattr(settings, opt.key, NO_VALUE)
+        if val is not NO_VALUE:
+            return opt.parser(val)
+
+        if opt.default is not NO_VALUE:
+            return opt.parser(opt.default)
+        raise ConfigurationError('"%s" must be configured in settings. See docs.' % opt.key)
```
willkg: Mmm... We can tweak this message and include the `help_text`. Plus the `opt.parser` calls should be wrapped in try/except with similar error handling.

standup/auth0/settings.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def __getattr__(self, key):
+        if key.startswith('AUTH0'):
+            return self.get(key)
+        super().__getattr__(self, key)
```
willkg: This makes the `app_settings.AUTH0_FOO` shorthand possible.

willkg:
@pmac I'm done all the coding things I wanted to do. I'll do a re-review tomorrow and probably fix some minor things, but otherwise this should be done now as much as we need at this juncture.

pmac:
Amazing! You're a hero among the people. I'll give it a final look in the morning. 🎉

bin/run-tests.sh
```diff
@@ -1,12 +1,13 @@
 #!/bin/bash -ex
 
+export DJANGO_SETTINGS_MODULE=standup.settings
```
willkg: Oops. This is from a bad merge. Bleh.

bin/run-tests.sh
```diff
@@ -1,12 +1,13 @@
...
 export DATABASE_URL=sqlite://
 export SECRET_KEY=itsasekrit
 export DJANGO_SETTINGS_MODULE=standup.settings
 export STATICFILES_STORAGE=django.contrib.staticfiles.storage.StaticFilesStorage
 
 export AUTH0_CLIENT_ID=ou812
-export AUTH0_CLIENT_SECRET=ou812
+export AUTH0_CLIENT_SECRET=secret_ou812
```
willkg: I changed this so they're different values so as to make sure the token request was correct.

standup/auth0/context_processors.py
```diff
@@ -1,32 +1,44 @@
...
-        AUTH0_CALLBACK_URL=quote_plus(settings.AUTH0_CALLBACK_URL),
-        # FIXME(willkg): This should be a token that ties both ends.
-        STATE='foo',
-    )
+class Auth0URL:
+    """Class that generates an Auth0URL on demand"""
+    def __init__(self, request):
+        self.request = request
+
+    def __str__(self):
```
willkg: This is nice because one of the side effects of this is that it creates a session. We don't want to create sessions for anyone who hits a page on the site.

standup/auth0/middleware.py
```diff
@@ -19,20 +21,20 @@ class Auth0LookupError(Exception):
 
 
 def renew_id_token(id_token):
-    """Renews id token and returns id token or None
+    """Renews id token and returns delegation result or None
 
     :arg str id_token: the id token to renew
 
-    :returns: ``id_token`` or ``None``
+    :returns: delegation result (dict) ``None``
```
willkg: This should read `"or ``None``".`

standup/auth0/models.py
```diff
@@ -3,11 +3,11 @@
...
+    """Associates a user with their id token
 
-    These tokens need to be renewed on a periodic basis and if they fail, they cause the Django
-    session to expire.
+    These tokens need to be renewed on a periodic basis and if they fail, they force Django logout.
 
     """
     user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
-    id_token = models.TextField(max_length=100, null=True)
+    id_token = models.TextField(help_text='The id token for this user.')
```
willkg: I couldn't find anything that denoted the sizing of this value. As far as I can tell, it can be arbitrary and thus we shouldn't cap it in the db.

standup/auth0/settings.py
```diff
@@ -0,0 +1,212 @@
...
+            msg = '%s\n\n%s' % (msg, opt.help_text)
+        raise ConfigurationError(msg)
+
+    def __getattr__(self, key):
+        if key.startswith('AUTH0'):
+            return self.get(key)
+        super().__getattr__(self, key)
+
+
+app_settings = AppSettings()
```
willkg: This is my new favorite way to do settings in Django libraries. It looks uninteresting here, but it's trivial to write a script that runs through the options and generates reST doc for Sphinx documentation with index points, formatting, etc. That makes it a lot easier to make sure docs match the code.

Plus it will throw helpful configuration errors. For example:

```
[13/Jan/2017 15:08:42] ERROR django.request: Internal Server Error: /statistics/
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/app/standup/status/views.py", line 319, in statistics
    'statsitems': groups,
  File "/usr/local/lib/python3.5/dist-packages/django/shortcuts.py", line 67, in render
    template_name, context, request=request, using=using)
  File "/usr/local/lib/python3.5/dist-packages/django/template/loader.py", line 99, in render_to_string
    return template.render(context, request)
  File "/usr/local/lib/python3.5/dist-packages/django_jinja/backend.py", line 84, in render
    context.update(processor(request))
  File "/app/standup/auth0/context_processors.py", line 40, in auth0
    'auth0configured': is_auth0_configured(),
  File "/app/standup/auth0/context_processors.py", line 9, in is_auth0_configured
    app_settings.AUTH0_CLIENT_ID and
  File "/app/standup/auth0/settings.py", line 215, in __getattr__
    return self.get(key)
  File "/app/standup/auth0/settings.py", line 211, in get
    raise ConfigurationError(msg)
standup.auth0.settings.ConfigurationError: "AUTH0_CLIENT_ID" must be configured in settings.

Your Auth0 client id. This is public-ish.
```

standup/auth0/models.py
pmac: I assume we'll have to lookup entries based on this field though, and some DB engines have trouble with indexing a text blob field. I think we should have this be a `CharField` with a large max_length, maybe 200, and set `db_index=True`?

standup/auth0/settings.py
pmac: Wow. Fancy.

standup/auth0/models.py
willkg: Mmm... I don't think we ever do that. As the code stands now, we query that table here and that's it:

* https://github.com/mozilla/standup/pull/325/files#diff-a1cf37ebb60cbdd2c99767226e989f5dR90
* https://github.com/mozilla/standup/pull/325/files#diff-6776186224a7b049a218c30b0cf2f46eR196

The only place where we query by `id_token` is in the tests.

So I think this is ok as it stands now, but I'm glad you mentioned it and we double-checked.

willkg:
Ok... I'm going to land this and test it on -stage. Here we go!

PR #327, opened 2017-01-11
Add pylibmc and support for Memcachier

pmac:


standup/settings.py
```diff
@@ -141,16 +141,35 @@ def path(*paths):
...
+else:
+    CACHES = {
+        'default': config('CACHE_URL', parser=django_cache_url.parse, default='locmem:default')
+    }
+
 if CACHES['default'].get('BACKEND', '') == 'django.core.cache.backends.locmem.LocMemCache':
     # If we're using locmem, set timeout for 60 seconds and restrict
     # cache to 100 things so as to not run rampant with memory usage.
     CACHES['default']['timeout'] = 60
     CACHES['default'].setdefault('OPTIONS', {})['MAX_ENTRIES'] = 100
```
willkg: Pretty sure we can ditch this `if-clause`. I added it when we thought cache was related to our memory problems and that turned out not to be the case.

willkg:
This looks fine to me. r+

PR #328, opened 2017-01-12
Cosmetic: clean up tests and requirements file

willkg:
* moves things around in requirements file making it more convenient to just
  do "hashin blah requirements.txt" and have the thing be in the right section
* remove things from setup.cfg that weren't doing anything
* centralize settings-related configuration into bin/run-tests.sh
* switch tests from using dummy cache to locmem cache which requires we clear
  the cache before every test

### DIFF
--- a/bin/run-tests.sh
+++ b/bin/run-tests.sh
@@ -2,10 +2,12 @@
 
 export DATABASE_URL=sqlite://
 export SECRET_KEY=itsasekrit
-# export AUTH0_CLIENT_ID=foo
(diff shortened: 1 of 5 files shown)


willkg:
This extracts some minor cleanup I did in #325 that isn't directly related to the auth0 work. It'll make that PR a little easier to review.

I'll wait for Travis to be happy, then merge this.

willkg:
Travis is happy. Looks good enough, so I'm going to merge it.

PR #330, opened 2017-01-12
Make github_handle optional and drop unique requirement

### LINKED ISSUE #329: nix requirements for github_handle field
We're not using Github auth anymore, so we no longer require someone to fill out the `github_handle` field in their profile and we don't need it to be unique.

We can probably drop the field altogether, but I suspect it'll help us merge accounts as we're settling into the new Auth0 system. So for now, I just want to make it optional and also drop the uniqueness requirement.

willkg:
Fixes #329

### DIFF
--- a/standup/status/migrations/0008_github_handle_fixes.py
+++ b/standup/status/migrations/0008_github_handle_fixes.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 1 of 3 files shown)


willkg:
Plus this fixes a problem I had with the profile page where it was caching things and thus information would come and go incorrectly.

willkg:
I think this looks good. Merging it.

PR #332, opened 2017-01-18
Add monkeypatch to prevent connection closing for memcachier

willkg:
This is mentioned at the bottom of the Django section here:

https://devcenter.heroku.com/articles/memcachier#django

> Finally, we also strongly recommend that you place the following code in your wsgi.py file to correct a serious performance bug (#11331) with Django and memcached. The fix enables persistent connections under Django, which by default uses a new connection for each request:
>
> ```
> # Fix django closing connection to MemCachier after every request (#11331)
> from django.core.cache.backends.memcached import BaseMemcachedCache
> BaseMemcachedCache.close = lambda self, **kwargs: None
> ```

### DIFF
--- a/standup/wsgi.py
+++ b/standup/wsgi.py
@@ -9,10 +9,15 @@ https://docs.djangoproject.com/en/1.8/howto/deployment/wsgi/
 
 import os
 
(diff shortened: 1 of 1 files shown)


willkg:
I'm going to merge this to see what happens.

PR #333, opened 2017-01-18
Install pylibmc via pip instead of apt

pmac:


willkg:
Oo--I like this possibility! I'll wait for Travis, then land it.

Issue #334, opened 2017-01-18
caching failures


willkg:
Standup -stage has caching enabled and using the Memcachier Heroku addon. However, we're seeing this in the logs:

```
2017-01-18T22:20:04.755231+00:00 app[web.1]: [18/Jan/2017 22:20:04] ERROR django.pylibmc:
MemcachedError: error 40 from memcached_get(:1:views.decorators.cache.cache_): (0x2dd1650)
FAILED TO SEND AUTHENTICATION TO SERVER, no mechanism available,  host: xxx:11211 ->
libmemcached/sasl.cc:238
```

Looking at the Memcachier dashboard, we see 0mb is used in the cache. I think both of these observations suggests that it's not working.

This issue covers fixing that.

willkg:
Today we looked at the configuration and though we didn't see any obvious issues, we redid our configuration based on the Memcachier docs:

https://devcenter.heroku.com/articles/memcachier#django

That didn't change anything.

Some searching found this:

* https://github.com/heroku/heroku-buildpack-python/issues/201
* http://stackoverflow.com/questions/27837326/memcachier-add-on-for-cloudcontrol-is-failing

That suggests maybe the problem is with our libmemcached.

@pmac tried switching from the packaged version of pylibmc to one we built ourselves during the deploy and that didn't change anything, either.

I think the next step is to look at building libmemcached. Maybe we've got an old version? Maybe the version we've got is bad in some way? Something like that.

willkg:
I spent some more time with it just now. Seems like we're hitting this:

https://bugs.launchpad.net/ubuntu/+source/libmemcached/+bug/1573594

I don't understand how this seems like a big deal, but there's like zip when doing a search for the error string on Google.

Anyhow, I think we have the following things we could try:

1. Switch to python-memcache pure python lib. It might not support auth and thus memcachier.
2. Switch to a different memcache service on Heroku.
3. Isolate this into a smaller project so we can debug/tinker with it more easily--should be straight-forward.
4. Run memcached in the container with our code. (@willkg isn't enthused about this option.)
5. Give up and switch to the db caching backend.

I'll try the first two tomorrow.

pmac:
We could also try redis as a cache server. That protocol is well supported and Heroku has an official plugin for it. Looks like neither an external memcache nor redis would be a secure place to store anything sensitive though.

https://devcenter.heroku.com/articles/securing-heroku-redis

willkg:
We're going to make a bunch of changes:

[...]

PR #336, opened 2017-01-19
Cache reworking

willkg:
This fixes everything for issue #334:

1. removes all caching and page caching
2. switches us back to using locmem
3. removes all the memcached stuff
4. changes auth0 `id_token` renewal to use the db and not cache--but also adds an untested cache-based middleware

@pmac Quick r?


willkg:
Thank you!

PR #339, opened 2017-01-19
Add rough finduser command

willkg:
This command looks at a bunch of User and StandupUser fields for a string and
then spits out a table of information for all the users that match in some way.

It's totally not wildly interesting, but it'll probably help a bunch plus it gives us something to iterate on going forward to make our lives easier.


willkg:
Good enough and doesn't affect Standup functionality, so I'm merging it in.

PR #340, opened 2017-01-20
Merge cli

willkg:
Add merge cli.

I'm pretty sure this does the right thing in most/all situations. However, since it's deleting the User, StandupUser and other related things, it's lossy, but it only loses user/standupuser information which is probably easy to re-enter.

### DIFF
--- a/standup/status/management/commands/mergeuser.py
+++ b/standup/status/management/commands/mergeuser.py
@@ -1,0 +1,85 @@
+from django.contrib.auth import get_user_model
+from django.core.management.base import BaseCommand
+from standup.status.models import StandupUser
+
+User = get_user_model()
+
+
+class Command(BaseCommand):
+    help = 'Merge two user accounts'
+
+    def add_arguments(self, parser):
+        parser.add_argument('--keep', type=int)
+        parser.add_argument('--delete', type=int)
+        parser.add_argument('-y', '--yes', dest='assume_yes', action='store_true')
+
+    def handle(self, *args, **options):
+        try:
+            user_keep = User.objects.get(id=options['keep'])
+        except User.DoesNotExist:
+            self.stdout.write('User keep=%s does not exist. Exiting.' % options['keep'])
+            return 1
+
+        try:
+            user_delete = User.objects.get(id=options['delete'])
+        except User.DoesNotExist:
+            self.stdout.write('User delete=%s does not exist. Exiting.' % options['delete'])
+            return 1
+
+        def get_status_count(user):
+            try:
+                return StandupUser.objects.get(user=user).statuses.count()
+            except StandupUser.DoesNotExist:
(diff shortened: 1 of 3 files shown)


willkg:
@pmac Merging is lossy, so it's worth looking at. Does this look ok? I made the ux for it deliberately annoyingly explicit, so you have to type something like:

```
$ ./manage.py merge --keep=USERID --delete=USERID
<prints lots of stuff about keep and delete>
Continue?: Y/n
...
```

Those user ids are the ones you'd see from running `./manage.py finduser TEXT`.

Does that seem ok?

standup/status/management/commands/mergeuser.py
```diff
@@ -0,0 +1,85 @@
...
+        self.stdout.write('Keeping and merging into:')
+        self.stdout.write('    id:          %s' % user_keep.id)
+        self.stdout.write('    username:    %s' % user_keep.username)
+        self.stdout.write('    email:       %s' % user_keep.email)
+        self.stdout.write('    date_joined: %s' % user_keep.date_joined)
+        self.stdout.write('    statuses:    %s' % get_status_count(user_keep))
+        self.stdout.write('')
+
+        if not options['assume_yes']:
+            self.stdout.write('Continue?: Y/n')
```
pmac: NIT: When I see one of these and one of the letters is capitalized (the Y in this case) I assume that's the default if I hit [enter] w/o typing a letter. That doesn't appear to be how this would work. Not important; just an observation.

standup/status/management/commands/mergeuser.py
```diff
@@ -0,0 +1,85 @@
...
+            cont = input()
+            if cont.strip().lower() != 'y':
+                self.stdout.write('Exiting.')
+                return 1
+
+        # Transfer statuses and profile information
+        if not hasattr(user_keep, 'profile'):
```
pmac: Huh. I didn't know `hasattr` would work like this. I thought all models had the attribute whether there was a matching record on the other end or not. TIL.

standup/status/management/commands/mergeuser.py
willkg: Oh funky. I always thought it was an indicator that it's not case sensitive.

willkg:
The `--yes` makes it easier to test. Otherwise the tests balk at providing output and input and it's messy.

willkg:
Thank you!

standup/status/management/commands/mergeuser.py
willkg: Wow--that's surprisingly difficult to search for, but I found this:

http://ux.stackexchange.com/questions/53640/default-values-for-text-based-confirmation-prompt#53644

That suggests my interpretation is totally bogus. I'll fix it.

PR #346, opened 2017-02-06
First pass at search

### LINKED ISSUE #164: search
It'd be nice to be able to search for bug numbers, pull requests, tags or specific text and have it show all the relevant status reports.

willkg:
This uses a self-written tokenizer which doesn't support parens and other
things. It's not very good. But probably a good enough first pass to start
iterating with.

Fixed #164.

### DIFF
--- a/standup/status/jinja2/base.html
+++ b/standup/status/jinja2/base.html
@@ -65,70 +65,72 @@
                 </ul>
               </li>
             {% endif %}
-              <li>
-                {% if user.is_authenticated() %}
-                  <a href="{{ url('users.profile') }}" id="user-menu">{{ gravatar(user, 22) }}</a>
-                  <ul>
-                    <li><a href="{{ url('users.profile') }}">Profile</a></li>
-                    {% if user.is_staff -%}
-                      <li><a href="{{ url('admin:index') }}">Admin</a></li>
-                    {%- endif %}
-                    <li><a id="logout-link" href="#">Logout</a></li>
-                  </ul>
-                {% else %}
-                  <a href="{{ url('users.loginform') }}"">Sign in</a>
-                {% endif %}
-              </li>
-            </ul>
-          </nav>
-        </div>
-      </div>
-    </header>
-
-    <div id="main-content" class="container_12 cf">
-      <div id="main-notices" class="grid_12">
-        <div class="notice success">
-          We've switched the way we use Auth0. If you have problems, see
-          <a href="{{ settings.HELP_FAQ_URL }}">our FAQ</a>.
-        </div>
-        {% if messages %}
-          {% for message in messages %}
-            <div class="notice {{ message.tags }}">{{ message }}</div>
-          {% endfor %}
-        {% endif %}
-      </div>
-      {% block before_content %}
-        {% if user.is_authenticated() %}
-          {{ statusize() }}
-        {% endif %}
-      {% endblock %}
-      {% block content %}{% endblock %}
-      {% block after_content %}{% endblock %}
-    </div>
-
-    <footer>
-      <div class="container_12 cf">
-        <div class="grid_1">
-          <img src="{{ static('img/logo-grey.png') }}" alt="">
-        </div>
-        <div class="grid_5">
-          Standup is a neat little app that helps to organize and publish
-          asynchronous status updates.
-        </div>
-        <div class="grid_6">
-          <nav>
-            <ul>
-              {% if settings.HELP_FAQ_URL %}
-                <li><a href="{{ settings.HELP_FAQ_URL }}">Help &amp; FAQ</a></li>
+            <li>
+              <a href="{{ url('status.search') }}">Search</a>
+            </li>
+            <li>
+              {% if user.is_authenticated() %}
+                <a href="{{ url('users.profile') }}" id="user-menu">{{ gravatar(user, 22) }}</a>
+                <ul>
+                  <li><a href="{{ url('users.profile') }}">Profile</a></li>
+                  {% if user.is_staff -%}
+                    <li><a href="{{ url('admin:index') }}">Admin</a></li>
+                  {%- endif %}
+                  <li><a id="logout-link" href="#">Logout</a></li>
+                </ul>
+              {% else %}
+                <a href="{{ url('users.loginform') }}"">Sign in</a>
               {% endif %}
-              <li><a href="https://github.com/mozilla/standup/issues">Report a bug</a></li>
-              <li><a href="https://github.com/mozilla/standup">Contribute</a></li>
-            </ul>
-          </nav>
-        </div>
+            </li>
+          </ul>
+        </nav>
       </div>
-    </footer>
+    </div>
+  </header>
+
+  <div id="main-content" class="container_12 cf">
+    <div id="main-notices" class="grid_12">
+      <div class="notice success">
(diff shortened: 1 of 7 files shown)


willkg:
This supports some basic search things like quotes denoting phrases, AND and OR operators and escaping quotes. It's doing icontains which sort of works, but doesn't give you back scores so it's less of a "search" and more of a "filter". It doesn't handle mispellings or stemming or a variety of other things.

Examples of syntax:

* foo
* "foo bar"
* foo AND bar
* foo OR bar
* "foo bar" AND bat AND "don\'t stop"

It doesn't support parens. It's missing a NOT. It doesn't let you create clauses around other fields like "created' or "user".

We could add parens and a NOT, but it's probably the case we should switch to a parsing library. Alternatively, we could update to Django 1.10 and use the postgres SearchVector stuff (https://docs.djangoproject.com/en/1.10/ref/contrib/postgres/search/)--I think that'd work and possibly better.

Anyhow, this is probably a good enough first pass. Further, it's not all done in the view, so we can trivially use the same code for a status API endpoint.

standup/status/jinja2/base.html
```diff
@@ -65,70 +65,72 @@
...
+              {% else %}
+                <a href="{{ url('users.loginform') }}"">Sign in</a>
+              {% endif %}
+            </li>
+          </ul>
+        </nav>
       </div>
-    </header>
+    </div>
+  </header>
```
willkg: I added a Search link and then fixed indentation.

standup/status/jinja2/base.html
```diff
@@ -65,70 +65,72 @@
...
+            {% if settings.HELP_FAQ_URL %}
+              <li><a href="{{ settings.HELP_FAQ_URL }}">Help &amp; FAQ</a></li>
+            {% endif %}
+            <li><a href="https://github.com/mozilla/standup/issues">Report a bug</a></li>
+            <li><a href="https://github.com/mozilla/standup">Contribute</a></li>
+          </ul>
+        </nav>
+      </div>
+    </div>
+  </footer>
```
willkg: The rest of the changes in this file are just indentation fixing and removing a spurious `</div>` tag.

standup/status/search.py
```diff
@@ -0,0 +1,208 @@
...
+        if token:
+            token.append(u'"')
+        else:
+            tokens[-1] = tokens[-1] + u'"'
+
+    if token:
+        # Add last token
+        tokens.append(u''.join(token))
+
+    return tokens
```
willkg: This comes from Fjord verbatim.

standup/status/search.py
```diff
@@ -0,0 +1,208 @@
...
+    escape = False
+    for c in text:
+        if not escape and c == u'\\':
+            escape = True
+            continue
+
+        new_text.append(c)
+        escape = False
+
+    return u''.join(new_text)
```
willkg: This comes from Fjord verbatim.

standup/status/search.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+def build_match(field, token):
+    return Q(**{'%s__icontains' % field: unescape(token)})
+
+
+def build_match_phrase(field, token):
+    return Q(**{'%s__icontains' % field: unescape(token)})
```
willkg: I wasn't sure how else to build this other than with `icontains`. It means that the word has to appear exactly in the content field. It also means that it'll pick up substrings. So if you searched for "antenna", it'd pick up all standup status entries for "antenna", "antennas", "antenna-loadtest", etc. That might be fine, but something like "end" might not be.

The consequence of this is that this implementation of search will kind of suck.

Having said that, we don't have better options without doing a lot more work.

standup/status/search.py
```diff
@@ -0,0 +1,208 @@
...
+
+    while tokens and tokens[-1] not in (u'OR', u'AND'):
+        token = tokens.pop()
+
+        if token.startswith(u'"'):
+            clauses.append(build_match_phrase(field, token[1:-1]))
+        else:
+            clauses.append(build_match(field, token))
+
+    return clauses
```
willkg: This is similar to the Fjord one, except Fjord is targeting Elasticsearch, so it tokenizes differently.

standup/status/search.py
```diff
@@ -0,0 +1,208 @@
...
+        lhs.extend(rhs)
+        return build_or(lhs)
+    elif token == u'AND':
+        lhs.extend(rhs)
+        return build_and(lhs)
+
+    # Note: This probably will never get reached given the way
+    # parse_match slurps. If the code were changed, it's possible this
+    # might be triggerable.
+    raise ParseError('Not an oper token: {0}'.format(token))
```
willkg: This is the same as in Fjord.

standup/status/search.py
```diff
@@ -0,0 +1,208 @@
...
+    :arg field: the field we're querying on
+    :arg tokens: list of tokens to consume
+
+    :returns: list of clauses
+    """
+    match_clauses = parse_match(field, tokens)
+    if tokens:
+        return [parse_oper(field, match_clauses, tokens)]
+    return match_clauses
+
```
willkg: This is the same as in Fjord.

standup/status/search.py
```diff
@@ -0,0 +1,208 @@
...
+    try:
+        clauses = parse_query(field, tokens)
+    except ParseError:
+        return build_match(field, text)
+
+    if len(clauses) > 1:
+        return build_or(clauses)
+
+    return clauses[0]
```
willkg: This is either identical to Fjord or pretty close.

standup/status/static/less/style.less
```diff
@@ -774,3 +774,37 @@ input[type='text'],
...
+        color: #aaa;
+      }
+    }
+    input[type=submit] {
+      width: 19%;
+
+    }
+  }
+  // FIXME(willkg): do the button here, too
+}
```
willkg: I should remove this comment. Otherwise this is "ok", but could use some help from someone better at CSS and styling.

standup/status/tests/test_search.py
```diff
@@ -0,0 +1,107 @@
...
+        ),
+        (
+            'foo', u'foo\\\\bar',
+            Q(foo__icontains=u'foo\\bar')
+        ),
+    ])
+    def test_edge_cases(self, field, text, expected):
+        # NOTE(willkg): We repr the Q objects because otherwise they don't compare. I think this is
+        # ok since I think it's order-dependent.
+        assert repr(generate_query(field, text)) == repr(expected)
```
willkg: These tests are lifted from Fjord.

standup/status/views.py
```diff
@@ -124,6 +125,22 @@ def statusize(request):
...
 
+class SearchView(PaginateStatusesMixin, TemplateView):
+    template_name = 'status/search.html'
+
+    def get_status_queryset(self):
+        query = self.request.GET.get('query')
+        if not query:
+            return []
+
+        return Status.objects.filter(generate_query('content', query))
```
willkg: Because of the way this was implemented, we can trivially wrap an API around it. Yay!

pmac:
I'm fully on board. This is a great first-pass, especially considering that most of this code has been in production for a long while in fjord. Let's land this and iterate. I've not yet tried it locally, but I'd be happy with merging as is and giving it a good try on staging. Thanks Will!

willkg:
Merging!

PR #352, opened 2017-02-08
Disable update API

willkg:
The update API endpoint needs to be rewritten. This disables it until we do
that.


willkg:
@pmac said this looks super, so I'm merging it.

PR #357, opened 2017-02-09
Fix @ mentions

### LINKED ISSUE #356: need consistent @ handling
In some places, the @ displayed is the `StandupUser.slug` or the `StandupUser.user.username`. In other places, it's `StandupUser.user.username`. The code that converts @ mentions to HTML links in status posts only matches on `StandupUser.slug`.

This issue covers:

1. writing a data migration that makes sure all users have a slug
2. changing all the code that looks at `StandupUser.user.username` to stop doing that
3. fixing the Profile page so that it is clear that the slug field is both for URLs as well as @ mentions

willkg:
* add migration to make sure all users have a slug
* change @ mentions so that it always uses the slug rather than User.username
* remove all uses of User.username--we don't want to use this anymore
* add slug-editing to Profile page

Fixes #356.

### DIFF
--- a/standup/status/forms.py
+++ b/standup/status/forms.py
@@ -22,8 +22,9 @@ class ProfileForm(forms.ModelForm):
     # as them.
     class Meta:
         model = StandupUser
-        fields = ['name', 'irc_nick']
+        fields = ['name', 'slug', 'irc_nick']
         labels = {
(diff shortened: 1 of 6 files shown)


willkg:
Thank you!

PR #358, opened 2017-02-09
Switch from nickname-based username to a hash

### LINKED ISSUE #349: need better User.username setting
The new auth0 code sets the `User.username` field to whatever we get back from auth0. It's not great. This is problematic because the standup user can't change it on their profile page and it's the field that's used for matching `@` mentions.

This issue covers figuring out a better way to do this.

willkg:
This changes the auth0 code from using the `user_info['nickname']` field for
usernames and instead generating the username from hashing the email address
which we're using as a unique identifier. This fixes all kinds of problems.

The only part of this that affects Standup is the part where we're generating User instances with a fresh username. Otherwise we don't use the username anywhere including when matching OIDC identities to Standup identities.

Fixes #349.

### DIFF
--- a/standup/auth0/pipeline.py
+++ b/standup/auth0/pipeline.py
@@ -1,13 +1,15 @@
+import base64
 from collections import OrderedDict
 from datetime import timedelta
+import hashlib
 import logging
 
 from django.contrib import messages
 from django.contrib.auth import get_user_model, login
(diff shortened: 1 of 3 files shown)


willkg:
^^^ That fixes profile generation so it's using email rather than username.

willkg:
Merging!

PR #360, opened 2017-02-10
Remove content_html

### LINKED ISSUE #208: content_html field not used
The Status model has a `content` field and a `content_html` field. Currently the two fields have the exact same content. Seems like it was the intention for `content` to have content be the raw content and `content_html` to have rendered content, but that didn't happen.

@pmac and I talked about this a bit. It's possible for things to change such that rendered content becomes stale over time. We probably don't want to render it and cache it forever. It's not clear whether caching it really has much impact on running the site right now.

Given all that, I think we should nix the `content_html` field and fix any code that's using it. After we get some monitoring in place and see how the site runs, we can figure out whether it makes sense to implement caching of the rendered content.

willkg:
We weren't using it anywhere, so this is just removing it from the model.

This fixes #208.

### DIFF
--- a/standup/status/admin.py
+++ b/standup/status/admin.py
@@ -31,6 +31,5 @@ class StatusAdmin(admin.ModelAdmin):
     list_display = ['id', 'user', 'project', 'content']
     search_fields = ['content']
     # FIXME(willkg): We want to make reply_to go away and having it causes the page to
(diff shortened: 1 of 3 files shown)


willkg:
I'm going to self-r+ this because it's so benign.

PR #361, opened 2017-02-10
Implement really rough smoke tests

### LINKED ISSUE #275: write up smoketest suite
We want a suite of tests that perform a smoketest on a site at a specified URL. This way we can run the smoketests locally but also on a -stage or -prod server.

For now, we're thinking of just writing a harness that uses requests and does stuff because we could throw that together pretty fast right now.

Later we should probably switch to selenium.

willkg:
This is good enough to build off of.

Fixes #275.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 DOCKERCOMPOSE = $(shell which docker-compose)
 PG_DUMP_FILE ?= standup.dump
+SERVER_URL ?= "http://web:8000"
 
 default: help
 	@echo ""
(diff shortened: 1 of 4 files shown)


willkg:
@pmac Does this look good enough for now? Does this help you with anything you've been working on?

pmac:
Seems like a good idea, and definitely looks like a good starting point. I hope to get back to making our docker stuff better. I'll be doing that when I get done with the bedrock docker/jenkins stuff.

willkg:
Given that, I'm going to land it as is for now. Thanks!

PR #363, opened 2017-02-13
Add team-handling to mergeuser

### LINKED ISSUE #344: mergeuser command should handle team memberships
It's unlikely this will happen, but I suppose it's possible. Namely, when merging two user accounts, we should carry over and accumulate all the team memberships.

The thinking here is that it's possible someone could create an account xyz, then something happens and their email address changes and they create a new account. Instead of merging the new one to the old one, maybe it's easier data-wise to merge the old one to the new one. If we did that, then we lose any team memberships the user was in.

I can't think of any cases where losing team memberships is a good thing, so I'm inclined to say that mergeuser should always migrate team memberships to the keep account.

This issue covers implementing that.

willkg:
We're merging users that are on teams. This fixes mergeuser so that it handles
teams correctly.

Fixes #344.

### DIFF
--- a/standup/status/management/commands/mergeuser.py
+++ b/standup/status/management/commands/mergeuser.py
@@ -1,6 +1,9 @@
 from django.contrib.auth import get_user_model
 from django.core.management.base import BaseCommand
-from standup.status.models import StandupUser
+from django.db import IntegrityError, transaction
+
+from standup.status.models import StandupUser, TeamUser
+
 
 User = get_user_model()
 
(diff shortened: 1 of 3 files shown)


standup/status/management/commands/mergeuser.py
```diff
@@ -77,8 +88,19 @@ def get_status_count(user):
...
         user_keep.save()
 
-        # Delete.
-        # FIXME(willkg): Hopefully this cascades....
+        # Copy teams from delete -> keep
+        self.stdout.write('Copying teams from delete -> keep')
+        if hasattr(user_delete, 'profile'):
+            for team in user_delete.profile.teams.all():
+                try:
+                    with transaction.atomic():
```
willkg: Without this line, the transaction model django has gets all fussy. I think it's because I'm lazily figuring things out by seeing if it throws an `IntegrityError` or not, but I'm not positive that's the issue. The error message I get back is this:

> django.db.transaction.TransactionManagementError: An error occurred in the current transaction. You can't execute queries until the end of the 'atomic' block.

willkg:
@pmac Can you look at this when you get a chance?

willkg:
Thank you!

PR #365, opened 2017-02-13
Fix email check

### LINKED ISSUE #362: [traceback] KeyError: 'email_verified'
https://sentry.prod.mozaws.net/ELOE/standup-prod/issues/383063/

```
Traceback:
  File "django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "newrelic/hooks/framework_django.py", line 503, in wrapper
    return wrapped(*args, **kwargs)
  File "django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "newrelic/hooks/framework_django.py", line 885, in wrapper
    return wrapped(*args, **kwargs)
  File "django/views/generic/base.py", line 89, in dispatch
    return handler(request, *args, **kwargs)
  File "standup/auth0/views.py", line 99, in get
    result = run_pipeline(app_settings.AUTH0_PIPELINE, **kwargs)
  File "standup/auth0/pipeline.py", line 48, in run_pipeline
    result = func(**kwargs) or {}
  File "standup/auth0/pipeline.py", line 244, in mozilla_auth0_pipeline
    return run_pipeline(pipeline, **kwargs)
  File "standup/auth0/pipeline.py", line 48, in run_pipeline
    result = func(**kwargs) or {}
  File "standup/auth0/pipeline.py", line 145, in reject_unverified_email
    if not user_info['email_verified']:
KeyError: 'email_verified'
```

The `user_info` for this user has an `email`, but not an `email_verified`. Looks like the user used LDAP to log in.

Mozilla did auth0 work over the weekend, so one possibility is that things have changed and we need to update our code.

Another possibility is that not all `user_info`s have an `email_verified`, but I have no idea why that would be the case since all three providers we're using should have verified email addresses.

Maybe we should switch it to check for either `email` or `email_verified`?

willkg:
If email_verified isn't in user_info, treat it as if it were False.

Fixes #362.

### DIFF
--- a/standup/auth0/pipeline.py
+++ b/standup/auth0/pipeline.py
@@ -142,7 +142,7 @@ def reject_unverified_email(request, user_info, **kwargs):
     :returns: ``HttpResponseRedirect`` or None
 
     """
-    if not user_info['email_verified']:
+    if not user_info.get('email_verified', False):
         # If inactive, add message and redirect to login page
         messages.error(
             request,


willkg:
This is benign--self-r+ and landing.

Issue #266, opened 2016-10-03
implement team management


willkg:
With the previous Standup, we could manage teams via the command line on the server or via the standup irc bot which used the API v2 endpoints. During the rewrite, we nixed both of these, but added an admin page.

The admin page mostly works, but there are problems. Further, we want to allow users to manage their own teams.

This issue covers thinking about teams and how to implement management, then spinning off the required issues.

willkg:
We've got an explicit `TeamUser` model defined that makes it hard to use the Django admin. Fixing that would make it easier for admin to make team changes. I don't think we have an issue for that, yet. That's probably a good first step allowing us time to think about how we want to make it user-managed.

PR #370, opened 2017-02-28
Fix teams and users

willkg:
Previously, we had a TeamUser m2m model and that made lots of things hard
including the managing teams in the Django admin.

This does a meticulous set of migrations to copy that data to an ephemeral
table, nix that m2m model and recreate the relationship using Django magic.

### DIFF
--- a/standup/status/admin.py
+++ b/standup/status/admin.py
@@ -14,16 +14,15 @@ class ProjectAdmin(admin.ModelAdmin):
 
 @admin.register(Team)
 class TeamAdmin(admin.ModelAdmin):
-    fields = ['name', 'slug']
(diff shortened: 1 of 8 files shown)


willkg:
This alleviates some of the problems in #266, but doesn't fix it.

Also, we need to update Django to 1.10 in order for the horizontal multi-select thing to work. So this change isn't super duper helpful until we do that.

pmac:
I don't see any problems. If it works for you and Travis doesn't barf then I think we're good to go. Thanks a bunch Will!

willkg:
^^^ That adds team members to the page.

If this doesn't trigger a Travis build, then I'll create a new PR.

PR #371, opened 2017-02-28
Teamuser fix

willkg:
Retrying the pull request.


willkg:
Per https://www.traviscistatus.com/incidents/hmwq9yy5dh9d Travis is still not working. We'll have to wait not only so it tests this, but also because all the deploy stuff is done via Travis.

Issue #372, opened 2017-03-01
upgrade standup to django 1.10


willkg:
Django 1.10 has fixes to the Django admin that make it work with CSP. One of the things that's busted is the horizontal multi-select thingermcdooder. Having that broken makes managing teams for admin sadmad.

Thus, we should update to Django 1.10 because sanity is a scarce resource that must be used gently and protected.

willkg:
I have this in a branch. When Travis starts working, I'll start landing things with reckless abandon.

PR #373, opened 2017-03-01
Update to Django 1.10.5

willkg:
This also updates:

* pipeline
* whitenoise

Because I was hitting weird errors and decided I might as well update some stuff
and see if that magically makes things go away. Contrary to 20 years of
experience with doing this kind of solution, it actually worked this time.


willkg:
This fixes #372.

standup/auth0/pipeline.py
```diff
@@ -26,9 +26,13 @@ class Exhausted(Exception):
...
-    return base64.urlsafe_b64encode(
-        hashlib.sha1(smart_bytes(email)).digest()
-    ).rstrip(b'=')
+    return (
+        base64.urlsafe_b64encode(
+            hashlib.sha1(smart_bytes(email)).digest()
+        )
+        .rstrip(b'=')
+        .decode('utf-8')
+    )
```
willkg: This makes sure that it always returns a unicode rather than a bytes. Not sure why I thought it was ok for it to return bytes before but ew.

standup/auth0/tests/test_views.py
```diff
@@ -113,4 +113,4 @@ def test_success(self, settings, db, client, messages_catcher):
             )
 
             # Assert we get redirected to the front page
-            assert is_redirect_to(resp, 'http://testserver/')
+            assert is_redirect_to(resp, '/')
```
willkg: The HTTP 302 class thingy sets Location to just the path rather than the protocol, host, port and path like it did before. So I had to adjust all the checks accordingly.

standup/status/tests/test_commands.py
```diff
@@ -35,7 +35,7 @@ def test_find_user(db):
             'Searching for: jimbob',
             '-----',
             'user_id  username',
-            '1        jimbob',
+            '%-5d    jimbob' % user.id,
```
willkg: For some reason, the user id isn't always 1. I suspect it's because either I was kicking off the tests in a way that reused the db or something. Anyhow, this is a good fix to make the tests less fragile.

willkg:
@pmac Quick r?

pmac:
Looks fine to me. Did you mean to upgrade django-pipeline? You said so in the PR, but I don't see any changes in the requirements file.

willkg:
Bah. I screwed up. I "upgraded" pipeline to the same version by mistake.

willkg:
I tweaked the commit messages when squash and merging.

PR #374, opened 2017-03-01
Adjust StandupUser __str__ for better str-fu

willkg:
This makes it easier to differentiate between people in the user multi-select
boxes.


willkg:
Yay!

Issue #367, opened 2017-02-15
Consider truncating URLs in a less short way


nox:
Most of the links I write on standups get [truncated](https://www.standu.ps/user/nox/), is it really useful for the maximum displayed length to be that short?

pmac:
It's useful to have some truncation due to really long words breaking the CSS width constraints, but we could increase the max for that. Do you have a suggestion?

nox:
@pmac Given what you just said, what about trying to make it as long as possible while still fitting the box it's in?

nox:
Actually, what about just prettifying GitHub links to `user@repos#number`?

pmac:
we do already link to a PR if you're posting in a project with a github repo configured and you say `pr #1234` in your status. [e.g.](https://www.standu.ps/status/42157/)

pmac:
not saying I wouldn't be happy with a more general github URL pretifier, and probably one for bugzilla links. Not sure I have time to put together a PR for it, but would happily review one.

willkg:
I'm going to change this to "prettify github urls" since that's the underlying issue and not truncation.

willkg:
I think this is something that we want. Further, I think this (supporting prettifying specified urls (github, bugzilla, etc)) is something we could easily do in a Filter once I get Bleach 2.0 out the door. I'll keep this in mind while I'm working on Bleach.

willkg:
Bugzilla added this: https://bugzilla.mozilla.org/show_bug.cgi?id=1309112

I'd be up for ditching our existing "pull #xyz" handling in favor of more general github "user/repo#num" support. Maybe we could go through and migrate all the existing statuses, too.

PR #379, opened 2017-03-09
Update to Bleach 2.0

willkg:
This updates Standup to use Bleach 2.0 and html5lib 0.999999999 (9 9s). It
switches to reusing a Cleaner and Linker which is an easy performance gain. It
also adds the nofollow linkify callback so that all status message links are
marked as "nofollow".


willkg:
I was thinking we can turn most of the passes in htmlify into either one html5lib filter or multiple filters. The advantage to that is that they get to work on the token stream rather than a chunk of HTML which would make them more focused and possibly faster.

Additionally, we can trivially add some more linkify callbacks for converting url link text into shortened version--something asked for in issue #367.

Anyhow, Bleach 2.0 gives us a lot of easy enhancement possibilities.

willkg:
@pmac Want to skim this? It's pretty benign--there isn't a lot of change.

standup/status/models.py
```diff
@@ -30,6 +32,10 @@
 ])
 
 
+CLEANER = Cleaner(tags=[])
+LINKER = Linker(callbacks=[trim_urls, nofollow])
```
willkg: Pretty sure we should have been doing "nofollow" all along. So I took the time in this PR to add that.

pmac: Very much agreed.

PR #385, opened 2017-05-09
Fix #382: Upgrade Django to 1.11.1

### LINKED ISSUE #382: upgrade to django 1.11
https://docs.djangoproject.com/en/1.11/releases/1.11/

We should upgrade to Django 1.11. I have no idea if it's straight-forward or not.

1.11 is the next LTS, so we want to be using that. Plus it has a bunch of nice postgres things that might be helpful for improving search. Plus it has an ExtractWeek function and related things that might reduce a lot of our "weekly view" code.

pmac:


pmac:
Tests pass for me locally, and I didn't see anything that should cause us problems in the [1.11 release notes](https://docs.djangoproject.com/en/1.11/releases/1.11/). If you agree we can do more testing manually on stage.

willkg:
I'll get to this today for sure.

PR #388, opened 2017-05-19
Switch "make lint" to use test image

### LINKED ISSUE #387: "make lint" should use test image--not web
When you run `make lint`, it starts up the db because it's using the web image. It doesn't need the db.

We should switch it to use the test image.

willkg:
Fixes #387


willkg:
This is super benign, so I'm going to self-r+ it.

PR #389, opened 2017-05-19
fix "make restore-db"

willkg:


willkg:
@pmac Can you look at this? This fixes it for me.

willkg:
The problem I was having was that stdout from cat wasn't making it to a `docker-compose exec` thing correctly.

Pretty sure it's this problem: https://github.com/docker/compose/issues/3352

I worked around the issue using the thoughts in the comments there.

pmac:
This works for me too! Nice work! r+

PR #394, opened 2017-05-31
Overhaul Standup to use mozilla-django-oidc

### LINKED ISSUE #345: drop our auth0 implementation and switch to mozilla-django-oidc
I didn't really understand what what we were implementing was an oidc client. Now that that's clearer, we should look at dropping our 1600 lines of auth0 stuff in favor of libraries.

This set looks promising:

* https://pypi.python.org/pypi/oic/0.9.4.0
* https://pypi.python.org/pypi/django-oidc/0.1.4

Maybe this one?:

* https://pypi.python.org/pypi/django-auth-oidc/0.1.0

willkg:
This reworks Standup to use mozilla-django-oidc and drops our home-grown auth0
library.

Fixes #345


willkg:
This is a work-in-progress. I still have a bunch of things that we should fix in mozilla-django-oidc first and then they need to do a version release.

Then I need to go through this again and remove some of the goofy things I did.

willkg:
Running `make test` passes, but `make test-image` fails. I'm puzzled and looking into it.

standup/settings.py
```diff
@@ -235,16 +216,15 @@ def path(*paths):
 CSP_FONT_SRC = (
     "'self'",
 )
-CSP_FORM_ACTION = ("'self'", AUTH0_DOMAIN) if AUTH0_DOMAIN else ("'self'",)
-CSP_CONNECT_SRC = ("'self'", AUTH0_DOMAIN) if AUTH0_DOMAIN else ("'self'",)
+CSP_FORM_ACTION = ("'self'", SITE_URL) if SITE_URL else ("'self'",)
+CSP_CONNECT_SRC = ("'self'", SITE_URL) if SITE_URL else ("'self'",)
```
willkg: I have no idea if this is right because I'm not sure what these additional bits are for. It's entirely possible this is hold-over from the previous previous auth0 implementation where we were using the JS lock file sources from `AUTH0_DOMAIN`.

willkg: @pmac What do you think about this? ^^^

pmac: You must be right. Can't think of any reason that Auth0 would be involved in CSP with the current implementation. Removal +1.

willkg:
^^^ That nixes those bits in `settings.py`. I'll wait for Travis to pass and assume that means this is good as far as we can tell at this point.

Then either tomorrow or Monday, I'll work through updating the -stage SSO bits with a service-now ticket, update the configuration, land this, and all that. Then we can test this out on -stage and see how it goes.

pmac:
Excellent! Excited to see how this goes 🥇

willkg:
I updated the configuration on -stage and did a service-now request to update the information on the SSO. So all set to merge this.

PR #395, opened 2017-06-14
Switch dockerfile base to mozmeao/base:pythode-3.6-6.10

pmac:
Turns out using tini as the entrypoint does not work with Heroku's Docker container runtime. Also I am now more a fan of Debian images than Alpine.

This does move back to having NodeJS installed in the production container, and we could move back to not having that, but I don't see it as a big enough deal to bother with switching to that at present. If anyone feels differently please let me know.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,9 +33,7 @@ build-base:
 	touch .docker-build-base
 
 build: .docker-build-base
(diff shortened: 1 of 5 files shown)


willkg:
Everything in here looks good to me. I'm going to land it so as to fix -stage and then we can fix any issues that come up in a future PR.

PR #396, opened 2017-06-14
Update to mozilla-django-oidc 0.3.0 and fixes some issues

willkg:
This:

* updates mozilla-django-oidc from 0.2.0 to 0.3.0
* removes `SITE_URL` which we don't need anymore
* fixes logout which was broken


willkg:
@pmac Does this look ok?

PR #416, opened 2017-07-25
415 auth0 idtoken

willkg:
Best I can figure, we have to do this in two steps:

1. re-add the auth0 app, remove the models, create a "nuke it all" migration (that's in this PR)
2. push everything to -stage and -prod
3. remove the auth0 app again

Quick PR for this one?


willkg:
Regarding "initial migration", we should be fine because all the migration history and other things should still be in the tables for -stage and -prod.

willkg:
Thank you!

PR #424, opened 2017-09-12
Improve mergeuser command

willkg:
We're hitting cases where we need to merge things from a new account into an
existing account, but keep the existing account's email address. This makes it
possible to do that with mergeuser.


willkg:
Does this look ok?

Example invocation:

```
$ ./manage.py mergeuser --keep=123 --delete=456 --keep-email
```

PR #427, opened 2017-09-20
Make restoredb.sh more resilient

willkg:
If bin/restoredb.sh starts the db container, then it needs to pause a bit for
the db container to come up.


willkg:
This isn't going to work because it runs on the host. I'm going to just can it.

PR #428, opened 2017-09-21
Fix finduser to show profile information in User query

willkg:
This fixes the issue I was having earlier today where information seemed to show up and disappear for the same user id depending on the search terms I was using.


willkg:
I'm going to self-r+ this since it's just a command we use to fix user issues and not part of the site so it's low risk.

PR #432, opened 2017-10-15
fixes #391 - nix replies

### LINKED ISSUE #391: remove replies
Standup has the ability to reply to a post. It's tricky to do, though. You can only do it via the irc bot. There's no web interface. I forget if we removed the parts of the templates that show replies or not. Generally, it's a "feature", but I think there's enough missing about it that it could be called half-implemented.

As of yesterday's standup db dump, there are 108 replies out of 45,477 status posts.

Right now, I want to simplify Standup to the important features that people need to use and ditch all the half-completed features. If at some point after a long conversation and a spec we decide to bring back replies, that's great--we can do it then. But for now, I want to remove replies altogether.

This issue covers that work.

willkg:
This removes the `reply_to` field and all the related bits.


standup/status/migrations/0017_fix_replies.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def fix_replies(apps, schema_editor):
+    """Adds 'In reply to url: ' to beginning of all replies"""
+    Status = apps.get_model('status', 'Status')
+
+    replies = Status.objects.filter(reply_to__isnull=False)
+    for reply in replies:
+        content = reply.content
+        content = 'In reply to https://www.standu.ps/status/%d/:\n%s' % (
```
willkg: I'm not sure this is the best thing ever, but it's good enough. There aren't many replies, so even if this isn't the best thing we could do, it feels like a "meh--let's just move on" issue.

willkg:
Pretty sure this is ok. I'm going to merge it and test it on -stage.

PR #433, opened 2017-10-16
Change "profile" to "edit profile" and add "my statuses" in menu

### LINKED ISSUE #413: Feature request: link to profile page.
Once logged in it would be nice to have a link to https://www.standu.ps/user/shobson/ from the same drop down with "Profile" and "Logout". I would probably change Profile to go to /user/shobson and add a "Edit Profile" link to go to /accounts/profile/.

willkg:
Fixes #413


willkg:
This changes the menu so that "Profile" is now "Edit profile" which makes a lot more sense. Further, it adds a "My statuses" entry which didn't exist before. Makes it easier to find your statuses.

willkg:
Thank you!

Issue #236, opened 2016-09-23
add way to do site-wide announcements


willkg:
It'd be really handy to have a way to do site-wide announcements so that we can communicate with Standup users various things like upcoming maintenance windows, incoming meteors, swag, comic books, ...

I've rolled my own in previous projects. It's not a hard thing to implement.

Another alternative is use django-soapbox (https://github.com/ubernostrum/django-soapbox). Pretty sure MDN uses that.

Whatever we do, it'd be super nice to be able to expose messages in any kind of status-check API endpoint, too. That way any non-website client can get status info and alert people who aren't using the website.

a2sheppy:
Yes, MDN uses django-soapbox. I don't know much about how it's implemented, although I will say that the result is decent.

PR #435, opened 2017-10-16
Issue #236: Add broadcast messages via the admin

pmac:
Also add an API endpoint for the messages.

### DIFF
--- a/standup/api/views.py
+++ b/standup/api/views.py
@@ -11,7 +11,10 @@ from django.http import (
 from django.utils.encoding import force_str
 from django.utils.text import slugify
 from django.views.decorators.csrf import csrf_exempt
+from django.views.decorators.http import require_safe
 from django.views.generic import View
+
+from soapbox.models import Message
 
 from .models import SystemToken
 from standup.status.models import Project, Status, StandupUser
@@ -68,6 +71,12 @@ class HttpResponseJSON(HttpResponse):
 
         if cors:
             self['Access-Control-Allow-Origin'] = '*'
+
+
+@require_safe
+def site_messages(request):
+    messages = [m.message for m in Message.objects.match('/api/')]
+    return HttpResponseJSON(messages, cors=True)
 
 
 class APIView(View):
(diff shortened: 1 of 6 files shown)


pmac:
Tried all of the locally and it seems fine. Nothing fancy though. Does allow for URL specific messages, which is nice. And if you use a URL of `/api/` for a message it will only go to the API.

standup/api/views.py
```diff
@@ -70,6 +73,12 @@ def __init__(self, content, status=None, cors=False):
             self['Access-Control-Allow-Origin'] = '*'
 
 
+@require_safe
```
willkg: I never knew that existed. I wonder if we should use this for other API endpoints.

standup/status/jinja2/base.html
```diff
@@ -91,10 +91,9 @@
 
   <div id="main-content" class="container_12 cf">
     <div id="main-notices" class="grid_12">
-      <div class="notice success">
-        We've switched the way we use Auth0. If you have problems, see
-        <a href="{{ settings.HELP_FAQ_URL }}">our FAQ</a>.
-      </div>
+      {% for message in get_messages_for_page(request) %}
+        <div class="notice success">{{ message }}</div>
```
willkg: I used "success" because it was green and I didn't want to alarm anyone. I don't know if we want all our notices to be success ones. I don't know if it matters a whole lot, but if messages had other levels, that might be good to expose.

willkg: I take it back--I don't see anything about levels in django-soapbox.

pmac: I was hoping it had some concept of `tag` or `class` or something we could use as a CSS class as well. Could roll our own or fork soapbox for that, but I decided for now it's not worth it.

standup/api/views.py
pmac: I think the others allow POST, so they need to allow most HTTP verbs. This is a read-only endpoint however. I like `require_safe` over `require_GET` since it also specifically allows `HEAD` requests.

standup/status/jinja2/base.html
pmac: We could change it to `info`, which I think is blue, so might be more appropriate for announcements. I was just sticking to using what had been working for now.

pmac:
Great. Let's try it on stage then.

PR #458, opened 2018-04-07
Fix #457: Make textarea vertically resizable

### LINKED ISSUE #457: Make "what's your status?" textarea resizable
Steps to reproduce:
  1. login to standup
  2. open https://www.standu.ps/
  3. write more than 5 lines of status into the textbox

Actual result:
  the text overflows from the textarea and scrollbar appears.
  User have to scroll in order to see the entire status text, and also it's not possible to see
  the entire status text at once

Expected result:
  either:
   * the textarea is manually resizable and user can expand it to make the whole status text visible without scrolling
   * the textarea expands automatically to make the whole status text visible without scrolling

arai-a:
This makes the "what's your status?" textarea vertically resizable with a handle, with `resize` style.
The minimum and initial height is the current height.

This works on Firefox, Safari, and Chrome, but not on Edge.
https://developer.mozilla.org/en-US/docs/Web/CSS/resize
Given that it's optional enhancement feature, would it be fine?

The resize handle is partially hidden because the textarea has round border, but it's almost visible on above supported browsers (tested on macOS, debian with xfce, and windows 10)

I've first tried making it also horizontally resizable, but it has unexpected behavior because the textarea is aligned to the right.
So, if I grab the handle at the bottom-right of the textarea and drag it to right, the left side of the textarea grows to the left.
so I'd like to make it only vertically resizable.

### DIFF
--- a/standup/status/static/less/style.less
+++ b/standup/status/static/less/style.less
@@ -656,9 +656,10 @@ input[type='text'],
       float: right;
       font-family: Helvetica, Arial, sans-serif;
       height: 92px;
+      min-height: 92px;
       margin-bottom: 20px;
       padding: 12px 20px;
-      resize: none;
+      resize: vertical;
       width: 530px;
 
       &::-webkit-input-placeholder {


willkg:
This is awesome--thank you!

willkg:
I pushed this to -stage, double checked it (looked super!), then pushed it to -prod.

Again--thank you for doing this!

PR #463, opened 2018-05-18
Tweaks admin and moves errormenow and statistics there

willkg:
This overrides the admin `index.html` template to add arbitrary pages that
aren't tied to Django models.

This moves the `statistics` page and the `errormenow` pages there. Now we
can enhance the `statistics` page to add more bits to it.


willkg:
Thank you! Pretty psyched about this, though it makes upgrading Django a little harder. :(

PR #466, opened 2018-06-07
Add site-wide messaging

### LINKED ISSUE #444: add site-wide messaging
django-soapbox wasn't updated for Django 2.0, so I removed that. That was how we were doing site-wide messaging.

This issue covers adding a new way to do site-wide messaging. We can either write our own app or use a library. In this case, seems easier to just write our own app.

willkg:
Fixes #444


pmac:
flake8 fail b/c migrations with long lines :( silly flake8.

pmac:
apart from unhappy travis this looks great! Nice work 🍰

willkg:
Awesome--thank you!

PR #474, opened 2018-07-16
Update Django to 1.11.14 and disable pyup

willkg:
This updates Django to 1.11.14 which is the latest version.

This also disables pyup. This project is effectively inactive at this point.
We definitely want to update to fix security vulnerabilities, but otherwise
don't want to spend time on the monthly requirements-upgrading hamster wheel.


pmac (approved):
:shipit:

willkg:
Thank you!

Issue #478, opened 2018-08-30
Implement API to extract data


Ms2ger:
Apparently the API that existed to extract data disappeared in a rewrite some time ago. It would be valuable to implement a simple one before decommissioning.

@simonsapin: I guess you might be interested as well

SimonSapin:
In my case I’ve consistently entered data through IRC, so I’ve decided that the easiest way to extract my data is from IRC logs. But yes, if someone wants to build an API that returns for example everything for a given user in a big JSON (ideally without pagination), I’d be interested in using it.

a2sheppy:
I would be happy to see this, since I need this information.

pmac:
There is now a JSON endpoint per user. e.g.

https://www.standu.ps/user/pmac.json

This will give you a json blob of all statuses for the user. Please let me know if you get errors as I'm not sure how many statuses this will handle.

a2sheppy:
WFM with 5157 entries.

PR #479, opened 2018-08-30
Add per user JSON stauses feed

pmac:


pmac:
@willkg I've no idea if the server will withstand a user with a lot of statuses, but this is an attempt.

pmac:
Should fix #478

standup/status/models.py
```diff
@@ -252,16 +252,17 @@ def htmlify(self):
...
         data = OrderedDict()
         data['id'] = self.id
         data['created'] = self.created.isoformat()
-        if self.user:
-            data['user'] = self.user.dictify()
-        else:
-            data['user'] = None
+        if include_user:
+            if self.user:
+                data['user'] = self.user.dictify()
```
willkg, line 263: I double-checked what this returns and I think that's fine as long as StandupUser.name isn't email-address-like.

willkg:
This looks fine to me as long as `StandupUser.name` isn't email-address-like. r+!

Thank you for doing this!

standup/status/models.py
pmac: For this new view I have `include_user=False` so it doesn't put any user info into the feed. Since you're requesting the feed of a user I figured that info was implied.

pmac:
Great! This isn't including any of the user data at all due to calling `status.dictify(False)`. So I think we're okay. I'm going to merge and we can see how it looks on stage.

